PR #12, opened 2018-07-25
*BSD platforms

FlingeR:
If this pull is approved I will send another for spyder to include BSD in the missing places where only Linux is taken into account.

### DIFF
--- a/spyder_kernels/customize/spydercustomize.py
+++ b/spyder_kernels/customize/spydercustomize.py
@@ -183,7 +183,7 @@ except:
 # Add default filesystem encoding on Linux to avoid an error with
 # Matplotlib 1.5 in Python 2 (Fixes Issue 2793)
 #==============================================================================
-if PY2 and sys.platform.startswith('linux'):
+if PY2 and ('linux' in sys.platform or 'bsd' in sys.platform):
     def _getfilesystemencoding_wrapper():
         return 'utf-8'
 


ccordoba12:
We don't officially support BSD OSes and we don't have a way to run our tests on them, so I'm a bit reluctant to go down this slippery slope.

ccordoba12:
Besides, you need to move this PR to our `0.x` branch.

ccordoba12:
> Remarks like: "all installation problems you encounter are expected to be solved by you, so please don't ask us for help."

Please read carefully what [we exactly say](https://github.com/spyder-ide/spyder/releases):

> We also support pip, but please be aware that pip installations are for advanced users with good knowledge of all Spyder dependencies. Because of that, all installation problems you encounter are expected to be solved by you, so please don't ask us for help.

It's not about Windows, it's about pip. And we put that warning because pip doesn't have a [dependency solver](https://github.com/pypa/pip/issues/988), so it's very, very easy to break a custom Python installation with pip.

But that's not the case with conda, so that's why we decided to give support to Anaconda users only (unless you prove to be a very knowledgeable user with a very concrete and detailed bug report).

> We don't officially support BSD OSes

That's true, what else do you want me to say?

> Do you officially support Darwin ?

Sure, but that's easy to support because Travis has VMs with macOS.

PR #13, opened 2018-07-25
*BSD platforms

FlingeR:
If this pull is approved I will send another for spyder to include BSD in the missing places where only Linux is taken into account.


ccordoba12:
Sorry, we're not interested on this.

PR #15, opened 2018-08-01
PR: Separate startup run_lines with semicolon instead of comma

### LINKED ISSUE #14: Startup lines to split with semicolon instead of comma
## Issue
Currently I can't use a IPyhton startup lines like these in Sypder since the line is parsed with comma separation:

```python
from modual import func1, func2
```
or 
```python
import os, sys
```

These common python import structures don't work in startup lines. I actually thought it was broken until I read the note in the ipython config and thought about the implications of using comma separation.

## Solution

If we use semicolon to separate the lines like you can do in python then any valid python should be possible.

Then you could use:
```python
import os, sys; from modal import func1, func2
```

The fix could be as simple as changing this comma to a semicolon in this code, or the line could be executed directly since it is valid python.
 
https://github.com/spyder-ide/spyder-kernels/blob/9a49376ec01ad41af5d3522758e5b4bb30ce85a8/spyder_kernels/console/start.py#L102-L106

## Potential Problem

This is a compatibility concern since the startup lines from 3 would not work in 4, but it's a good time to make a braking change or we can tell that a config file is from 3 it can be upgraded to 4 by replacing the commas with semicolons automatically.

bcolsen:
Fixes #14 

needs a PR for Spyder config as well

### DIFF
--- a/spyder_kernels/console/start.py
+++ b/spyder_kernels/console/start.py
@@ -101,7 +101,7 @@ def kernel_config():
     # Run lines of code at startup
     run_lines_o = os.environ.get('SPY_RUN_LINES_O')
     if run_lines_o is not None:
-        spy_cfg.IPKernelApp.exec_lines = [x.strip() for x in run_lines_o.split(',')]
+        spy_cfg.IPKernelApp.exec_lines = [x.strip() for x in run_lines_o.split(';')]
     else:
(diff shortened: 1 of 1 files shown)


ccordoba12:
I'll merge after releasing a new version of `0.x`.

PR #17, opened 2018-08-04
PR: Add/update descriptions, links and metadata in setup.py

CAM-Gerlach:
Along the lines of spyder-ide/spyder#7582 as well as per our discussion with the Debian/Ubuntu maintainers (e.g. @ghisvail ), I've added a short and long description to the setup.py docstring and ``setup()`` call per the latter, and updated the other metadata and links following the former.


setup.py
```diff
@@ -6,10 +6,35 @@
...
+Spyder Kernels
 ==============
 
-Jupyter kernels for the Spyder console
+Provides Jupyter kernels for use with the Spyder IDE's IPython consoles,
+launched either through Spyder itself or in an independent Python session.
+These allow for interactive or file-based execution of Python code in
+different environments inside of Spyder.
+
+Spyder is a powerful scientific environment written in Python, for Python,
```
ccordoba12: This paragraph applies to Spyder but not to `spyder-kernels`. Please remove it.

setup.py
```diff
@@ -6,10 +6,35 @@
...
+and designed by and for scientists, engineers and data analysts.
+It features a unique combination of the advanced editing, analysis, debugging
+and profiling functionality of a comprehensive development tool with the data
+exploration, interactive execution, deep inspection and beautiful visualization
+capabilities of a scientific package. For more on Spyder, see our website:
+https://www.spyder-ide.org/
+
+To launch a Spyder IPython console and run code in a different environment
```
ccordoba12: This doesn't need to be part of the package description. Please remove it.

ccordoba12: It's very specific to how it works and it's going to change when we add our `ipdb` kernel.

setup.py
```diff
@@ -44,19 +70,30 @@ def get_version(module='spyder_kernels'):
 setup(
     name='spyder-kernels',
     version=get_version(),
-    keywords='spyder jupyter kernel',
+    keywords='spyder jupyter kernel ipython console shell',
```
ccordoba12: Please remove `shell` from here. This is not something like `Xonsh`.

setup.py
```diff
@@ -44,19 +70,30 @@ def get_version(module='spyder_kernels'):
...
     url='https://github.com/spyder-ide/spyder-kernels',
+    download_url="https://www.spyder-ide.org/#fh5co-download",
     license='MIT',
     author='Spyder Development Team',
-    description='Jupyter kernels for the Spyder console',
+    author_email="spyderlib@googlegroups.com",
+    description="Jupyter kernels for the Spyder IDE's IPython consoles.",
```
ccordoba12: The previous description is fine:

1. We only have one console, so there's no need for `IPython console`.
2. IDE is ugly. You don't see the Pycharm IDE, just Pycharm.

So, this should be:

    Jupyter kernels for the Spyder's console

setup.py
```diff
@@ -44,19 +70,30 @@ def get_version(module='spyder_kernels'):
...
                  'License :: OSI Approved :: MIT License',
                  'Operating System :: OS Independent',
+                 'Programming Language :: Python :: 2',
                  'Programming Language :: Python :: 2.7',
-                 'Programming Language :: Python :: 3']
+                 'Programming Language :: Python :: 3',
+                 'Programming Language :: Python :: 3.4',
+                 'Programming Language :: Python :: 3.5',
+                 'Programming Language :: Python :: 3.6',
+                 'Programming Language :: Python :: 3.7',
```
ccordoba12: I don't want to keep updating this list after each Python release. So please leave it was it was:

    'Programming Language :: Python :: 2.7',
    'Programming Language :: Python :: 3'

I borrowed this from the IPython/Jupyter guys, by the way.

setup.py
```diff
@@ -44,19 +70,30 @@ def get_version(module='spyder_kernels'):
...
     license='MIT',
     author='Spyder Development Team',
-    description='Jupyter kernels for the Spyder console',
+    author_email="spyderlib@googlegroups.com",
+    description="Jupyter kernels for the Spyder IDE's IPython consoles.",
+    long_description="\n".join(DOCLINES[4:]),
     packages=find_packages(exclude=['contrib', 'docs', 'tests*']),
     install_requires=REQUIREMENTS,
     include_package_data=True,
+    platforms=["any"],
```
ccordoba12: No, that's not true. Please remove this.

setup.py
CAM-Gerlach: Okay, removed.

setup.py
CAM-Gerlach: Sorry, removed.

setup.py
CAM-Gerlach: Both good points; I include "IDE" because (unlike PyCharm) not everyone knows what Spyder is, but I agree it looks ugly. Reverted to (fixed grammar):

```
Jupyter kernels for Spyder's console
```

setup.py
CAM-Gerlach: Okay; I was referencing the SciPy setup.py, incidentally (among others). 

I mean, its not *that* bad to update it (or not, as appropriate) once every one or two years when we add or drop compatibility with a Python version, since it has an actual purpose: to make it clearer to users browsing PyPI what we've compatible with (since we don't support all Python 3 versions, just >=3.4, and its not impossible we might temporarily be incompatible with a new release). Its also now inconsistent with the main Spyder setup.py since you approved these same changes there. But I reverted it back for now...

setup.py
CAM-Gerlach: Okay, removed. The reason I added it was because ``'Operating System :: OS Independent',`` was included as a trove classifier. Should this, then, also be removed, or does ``spyder-kernels`` fall into some edge case where one is true but not the other? (Also, as a sidenote, Spyder itself was marked ``"any"`` in its setup.py until my recent PR fixed it).

CAM-Gerlach:
@ccordoba12 Addressed each of your comments as indicated in the respective replies. Also ported the relevant changes over to conda-forge/spyder-kernels-feedstock#7 .

CAM-Gerlach:
@ccordoba12 Thanks! As a quick followup, I am genuinely curious on how can it be both ``Operating System :: OS Independent`` yet not ``platform = ["any"]``.

PR #19, opened 2018-08-06
PR: Fix inconsistent CRLF EOLs in __init__.py, py3compat.py and utils/misc.py

CAM-Gerlach:
All of the files in the repo are properly harmonized on LF (``0x0A``) EOLs, except for a few outliers: a few ``__init__,py``s for some reason (which are conveniently all empty/almost empty), ``utils/misc.py`` which is rather short, and ``py3compat.py``. Therefore, this PR standardizes those files on LF as well, so the entire repo has one consistent line break convention going forward. Aside from the maintenance and consistency benefits and conforming to the standard best practice, it also leads to much cleaner ``git diff`` output, 3-5% smaller file sizes for the same content, better consistency between Windows and *nix systems, and maximizes interoperability.

Obviously, the foremost concern with this change, even if relatively limited in scope, is merge conflicts and the like. However, that currently should be a non-issue, since no other open PRs touch any of these files, and they are each identical between ``0.x`` and ``master`` which should lead to a clean, conflict-free merge to the latter branch as well (which testing seems to confirm), so everything is harmonized.

Thanks!


CAM-Gerlach:
> Well thought @CAM-Gerlach! We should unify EOLs in new repos.

@ccordoba12 Excellent; really glad we're of the same mind.

> please create a file called .gitattributes

This is actually just what I did in the ``spyder-docs`` repo when I first created it: harmonized everything on LF and set up ``,gitattributes`` as you described, so I pulled over that ``.gitattributes`` as a starting point. I then probably went a *little* overboard explictly configuring it for every current or likely file type, to avoid any possibility of file corruption or missed inconsistent line endings due to git guessing wrong. However, the advantage is that we can use this same ``.gitattributes`` across all our repos with ease, since it should include every file type we are likely to encounter across any Spyder-related project, and thus we can port it over to ``website-spyder``, ``QtPy``, ``spyder-docs``, our plugins, etc. (and eventually the main repo itself, once we eventually break free of the CRLF curse there).

The one thing change from the above suggestion, as implemented in both the current ``.gitattributes`` on our ``spyder-docs`` repo and the one pushed here, is using ``* text=lf`` (with exceptions for Windows-specific files like ``.bat``), not ``* text=auto``, since the former will "maintain the same EOLs from now on" on all platforms and ensure the consistency and benefits mentioned above are present for all users, while the latter will actually make things worse for Windows users by trying to involuntarily convert *everything* (including files that should always stay LF) back to CRLF in their working tree, which furthermore for current users requires either a complicated series of steps or deleting their entire repository and starting fresh to prevent problems as the files are currently in their native line endings, i.e. LF except for those explicitly changed by this PR

Furthermore, this ensures consistent line endings across all platforms for all files, brings all the above-mentioned benefits of all-LF line endings to Windows users like myself, follows workflow best practices harmonizing on one line ending everywhere, avoids the overhead and complexity of constant conversion, makes the ``git diff`` output much cleaner (as it will constantly display spurious <kbd>^M</kbd> symbols on any changed lines), ensures the working tree is always consistent with the repo itself.

PR #41, opened 2018-08-28
PR: Use a mixin to share functionality between console and ipdb kernels

### LINKED ISSUE #26: Create BaseKernel with common elements between our kernels

dalthviz:
Fixes #26


spyder_kernels/base/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+# -----------------------------------------------------------------------------
+# Copyright (c) 2018- Spyder Kernels Contributors
+#
+# Licensed under the terms of the MIT License
+# (see spyder_kernels/__init__.py for details)
+# -----------------------------------------------------------------------------
+
+"""
+Base kernel
+"""
```
ccordoba12: We don't need this as a package because this is not really a kernel. So please remove this directory.

spyder_kernels/base/kernel.py
```diff
@@ -0,0 +1,372 @@
+# -*- coding: utf-8 -*-
```
ccordoba12: Please move this file to `spyder_kernels/base_kernel_mixin.py`

spyder_kernels/base/kernel.py
```diff
@@ -0,0 +1,372 @@
...
+# shown at all there)
+EXCLUDED_NAMES = ['In', 'Out', 'exit', 'get_ipython', 'quit']
+
+# To be able to get and set variables between Python 2 and 3
+PICKLE_PROTOCOL = 2
+
+PY2 = sys.version[0] == '2'
+
+
+class BaseKernelMixIn:
```
ccordoba12: Classes need to inherit from `object` to be new classes in Python 2, i.e.

```python
class BaseKernelMixIn(object):
```

ccordoba12:
This is looking really good @dalthviz, thanks! I left a couple of comments to better organize the code you introduced.

My only recommendation is for you to think what to do to avoid repeated tests for the functionality present in `BaseKernelMixin` when using `ConsoleKernel` and `IPdbKernel`. I think you could use a [parametrized fixture](https://docs.pytest.org/en/latest/fixture.html#parametrizing-fixtures) to achieve this.

spyder_kernels/base/kernel.py
ccordoba12: Or even better, let's call that file `kernelmixin.py`

spyder_kernels/tests/test_kernelmixin.py
```diff
@@ -0,0 +1,389 @@
...
+# Constants
+# =============================================================================
+FILES_PATH = os.path.dirname(os.path.realpath(__file__))
+
+
+# =============================================================================
+# Tests
+# =============================================================================
+# --- For the Variable Explorer
+def test_get_namespace_view(console_kernel, ipdb_kernel):
```
ccordoba12: @dalthviz, please avoid repeated code in these tests. For that I think you can create a single fixture called `create_kernel` parametrized with the two kernels we want to test: ConsoleKernel and IPdbKernel.

spyder_kernels/utils/test_utils.py
```diff
@@ -19,10 +19,10 @@
 except ImportError:
     from io import StringIO
 
-from spyder_kernels.console.kernel import SpyderKernel
+from spyder_kernels.console.kernel import ConsoleKernel
 
 
-def get_kernel(kernel_class=SpyderKernel):
+def get_kernel(kernel_class=ConsoleKernel):
```
ccordoba12: This function should be called now `get_console_kernel`.

spyder_kernels/ipdb/tests/test_ipdb_kernel.py
```diff
@@ -30,12 +30,48 @@
...
+                                                         'Infinity',
+                                                         'sctypes',
+                                                         'rcParams',
+                                                         'rcParamsDefault',
+                                                         'sctypeNA', 'typeNA',
+                                                         'False_', 'True_'],
+                                      'minmax': False}
+    # Teardown
+    def reset_kernel():
+        kernel.do_execute('reset -f', True)
```
ccordoba12: Is the `%reset` magic defined in MetaKernel? I think it's not.

spyder_kernels/utils/test_utils.py
ccordoba12: Or does it work for both kernels?

dalthviz: Works for both. As a parameter you give the class of the kernel you want to test.

ccordoba12: Ok, I see.

spyder_kernels/tests/test_kernelmixin.py
```diff
@@ -0,0 +1,267 @@
...
+
+
+# =============================================================================
+# Fixtures
+# =============================================================================
+@pytest.fixture(scope="module",
+                params=[IPdbKernel,
+                        ConsoleKernel])
+def kernel(request):
+    """IPdb kernel fixture"""
```
ccordoba12: This needs to be `Kernel fixture`.

ccordoba12:
> I think this is ready

I think so too. Thanks a lot @dalthviz! This is great work!

Please don't add more commits here. I'm going to do a bit of squashing and refactoring, and then I'll merge.

PR #49, opened 2018-09-22
PR: Make do_inspect and help work as in IPython

### LINKED ISSUE #39: do_inspect is failing with IPdb kernel
Right is throwing an error every time a parenthesis is open in Qtconsole.

dalthviz:
Fixes #39

### DIFF
--- a/spyder_kernels/ipdb/kernel.py
+++ b/spyder_kernels/ipdb/kernel.py
@@ -22,6 +22,7 @@ from IPython.core.completer import IPCompleter
 from IPython.core.inputsplitter import IPythonInputSplitter
 from IPython.core.interactiveshell import InteractiveShell
 from IPython.core.debugger import BdbQuit_excepthook
+from IPython.utils.tokenutil import token_at_cursor
 from metakernel import MetaKernel
 
 from spyder_kernels._version import __version__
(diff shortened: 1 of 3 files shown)


spyder_kernels/ipdb/kernel.py
```diff
@@ -184,8 +184,11 @@ def get_usage(self):
 
     def get_kernel_help_on(self, info, level=0, none_on_fail=False):
         """Help for Pdb commands."""
-        cmd = info['code'].strip()
-        return self.debugger.do_help(cmd)
+        if none_on_fail:
+            return None
+        else:
+            cmd = info['code'].strip()
+            return self.debugger.do_help(cmd)
```
ccordoba12: This avoid the error I reported and it's nice. But what I really want is to show the same help as it's shown by qtconsole when it's using ipykernel.

And according to this

https://github.com/ipython/ipykernel/blob/e32d7f9360e345e87bc4f10dbb66831421b0900b/ipykernel/ipkernel.py#L401

it shouldn't be too hard to implement, now that we have an `ipyshell` object in the IPdb kernel.

dalthviz:
After testing the help you get with `?` only works for the commands that you can list with the debugger help:

imagen

spyder_kernels/ipdb/kernel.py
```diff
@@ -184,8 +206,26 @@ def get_usage(self):
...
+                except AttributeError:
+                    getattr(self.debugger, 'do_' + cmd)
+                    help_on = self.debugger.do_help(cmd)
+            except AttributeError:
+                response = self.do_inspect(info['code'], info['help_pos'])
+                if 'data' in response:
+                    if 'text/plain' in response['data']:
+                        help_on = response['data']['text/plain']
+                if not help_on:
+                    self.debugger.error('No help for %r' % cmd)
```
ccordoba12: @dalthviz, the test failure is caused because `error` doesn't exist in Pdb for Python 2.

However, you can add it to our SpyderPdb class so that it works for Python 2 and 3.

This is how it's defined in Python 3:

https://github.com/python/cpython/blob/3.6/Lib/pdb.py#L455

PR #51, opened 2018-09-23
PR: Override 'do_complete' method to prevent duplicated path matches

### LINKED ISSUE #50: Repeated path completions
paths

dalthviz:
Fixes #50

### DIFF
--- a/spyder_kernels/ipdb/kernel.py
+++ b/spyder_kernels/ipdb/kernel.py
@@ -64,7 +64,10 @@ class DummyShell(object):
 
 
 class IPdbCompleter(IPCompleter):
-    """Subclass of IPCompleter without file completions."""
+    """
+    Subclass of IPCompleter without file completions so they don't
+    interfere with the ones provided by MetaKernel.
+    """
 
     def file_matches(self, text):
         """Return and empty list to deactivate file matches."""
--- a/spyder_kernels/ipdb/kernel.py
+++ b/spyder_kernels/ipdb/kernel.py
@@ -63,6 +63,14 @@ class DummyShell(object):
         return DummyMagicsManager()
 
 
+class IPdbCompleter(IPCompleter):
+    """Subclass of IPCompleter without file completions."""
+
+    def file_matches(self, text):
+        """Return and empty list to deactivate file matches."""
+        return []
+
+
 class IPdbKernel(BaseKernelMixIn, MetaKernel):
     implementation = "IPdb Kernel"
     implementation_version = __version__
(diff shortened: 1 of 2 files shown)


spyder_kernels/ipdb/kernel.py
```diff
@@ -162,6 +162,71 @@ def do_is_complete(self, code):
             r['indent'] = ' ' * indent_spaces
         return r
 
+    def do_complete(self, code, cursor_pos):
```
ccordoba12: Did you copy this method from MetaKernel? If that's the case, what if we submit a PR to them instead to disable their `path_matches` and only use our own?

I think that should simpler than this.

dalthviz: Yep, I took the same method although what I did was to do a replacement of the `"` for each element in `info['path_matches']` to prevent duplicated entries when extending the matches with the ones returned by `get_completions` (validating if the `get_completions` matches are the same as the ones that Metakernel retrieves).

Should I do a PR against Metakernel then?

ccordoba12: Yes, I think it's better. If we can't use `super(Metakernel, self).do_complete(...)` at some point in this re-implementation, then that's the way to go to maintain our own implementation of `do_complete`.

ccordoba12:
@dalthviz, a test is missing for this.

spyder_kernels/ipdb/tests/test_ipdb_kernel.py
```diff
@@ -231,6 +231,13 @@ def test_ls_path_complete(ipdb_kernel):
         assert comp['matches'] == ['"ipython\\"'], comp
 
 
+def test_no_path(ipdb_kernel):
+    kernel = ipdb_kernel
+    info = {'code': '~/.ipytho'}
```
ccordoba12: @dalthviz, let's better get completions for `/tm` and assert that there's only one completion returned.

Also, skip this test for Windows (given that `/tmp` only exists on Linux).

spyder_kernels/ipdb/tests/test_ipdb_kernel.py
```diff
@@ -231,6 +231,13 @@ def test_ls_path_complete(ipdb_kernel):
         assert comp['matches'] == ['"ipython\\"'], comp
 
 
+@pytest.mark.skipif(os.name == 'nt', reason="Only for linux")
```
ccordoba12: Please change this to

    not sys.platform.startswith('linux'), reason="Only works on Linux"

because I don't know if `/tmp` is present in macOS

spyder_kernels/ipdb/kernel.py
```diff
@@ -63,6 +63,14 @@ def lsmagic(self):
         return DummyMagicsManager()
 
 
+class IPdbCompleter(IPCompleter):
+    """Subclass of IPCompleter without file completions."""
```
ccordoba12: Please change this docstring to

```python
"""
Subclass of IPCompleter without file completions so they don't
interfere with the ones provided by MetaKernel.
"""
```

PR #77, opened 2019-01-17
PR: Have runcell set __file__ to the path of the file containing the cell 

### LINKED ISSUE #76: Detect the name of the file currently running the cell
I think it would be useful to obtain the name of the file currently running the cell.

__spyder_cell_file__

It would be nice if I could get to the point in my script, that would programmatically:
 * Save the current working file
 * Copy it to some known directory for archival purposes

Checking for the existance of the `__file__` is a little problematic, because it can be set by a `run` command, and it won't be unset.

The current working directory if often clobbered when restarting kernels and navigating through the file explorer.

XREF: https://stackoverflow.com/questions/54221512/get-name-of-file-where-the-cell-is-running-from-in-spyder/54224444#54224444

bcolsen:
Fixes #76

### DIFF
--- a/spyder_kernels/console/tests/test_console_kernel.py
+++ b/spyder_kernels/console/tests/test_console_kernel.py
@@ -376,7 +376,7 @@ def test_runcell(tmpdir):
 
     with setup_kernel(cmd) as client:
         # Write code with a cell to a file
-        code = u"result = 10"
+        code = "result = 10;print(__file__)"
         p = tmpdir.join("cell-test.py")
         p.write(code)
 
@@ -387,7 +387,9 @@ def test_runcell(tmpdir):
 
         # Execute runcell
         client.execute(u"runcell('', r'{}')".format(to_text_string(p)))
-        client.get_shell_msg(block=True, timeout=TIMEOUT)
+        msg = client.get_shell_msg(block=True, timeout=TIMEOUT)
+        print(msg['content'])
+        assert "cell-tests.py" in msg['content']
 
         # Verify that the `result` variable is defined
         client.inspect('result')
@@ -395,6 +397,14 @@ def test_runcell(tmpdir):
         print(msg['content'])
         content = msg['content']
         assert content['found']
+
+        # Verify that the `__file__` variable is undefined
+        client.inspect('__file__')
+        msg = client.get_shell_msg(block=True, timeout=TIMEOUT)
+        print(msg['content'])
+        content = msg['content']
+        assert not content['found']
+
 
 def test_np_threshold(kernel):
     """Test that setting Numpy threshold doesn't make the Variable Explorer slow."""
--- a/spyder_kernels/console/tests/test_console_kernel.py
+++ b/spyder_kernels/console/tests/test_console_kernel.py
@@ -392,21 +392,18 @@ def test_runcell(tmpdir):
         # Verify that the `result` variable is defined
         client.inspect('result')
         msg = client.get_shell_msg(block=True, timeout=TIMEOUT)
-        print(msg['content'])
         content = msg['content']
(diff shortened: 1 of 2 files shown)


spyder_kernels/customize/spydercustomize.py
```diff
@@ -753,6 +755,7 @@ def runcell(cellname, filename):
     ipython_shell.events.trigger('post_execute')
 
     ipython_shell.run_cell(cell_code)
+    namespace.pop('__file__')
```
hmaarrfk: So __file__ only lives during the execution of the cell, not afterwards?


Just an FYI, this is contrary to the "run file" functionality, at least in 3.x

hmaarrfk: Oh you are right, maybe I was running an older version.
I think it is ok to remove `__file__`, probably for the best anyway.

ccordoba12: Yes, I also think removing `__file__` after the command finishes is the best thing to do.

spyder_kernels/customize/spydercustomize.py
```diff
@@ -740,6 +740,8 @@ def runcell(cellname, filename):
         pass
     run_umr()
     ipython_shell = get_ipython()
+    namespace = _get_globals()
```
bcolsen: @ccordoba12 Should I just use `global __file__` here instead? If that works. I haven't tested it.

ccordoba12: No. I think the way you implemented this is the right way to do it.

spyder_kernels/console/tests/test_console_kernel.py
```diff
@@ -387,7 +387,9 @@ def test_runcell(tmpdir):
 
         # Execute runcell
         client.execute(u"runcell('', r'{}')".format(to_text_string(p)))
-        client.get_shell_msg(block=True, timeout=TIMEOUT)
+        msg = client.get_shell_msg(block=True, timeout=TIMEOUT)
+        print(msg['content'])
+        assert "cell-tests.py" in msg['content']
```
bcolsen: @ccordoba12 I can't get the output from the print statement in the code. Any suggestions? 

I can't use inspect because I remove the definition. If I don't remove it and then inspect I get the filename of the test.

ccordoba12: Instead of printing `__file__`, please assign the contents of `__file__` to a variable, then inspect its contents. That should be enough.

bcolsen: Thank you! That works of course. I was too laser focused on trying to check `__file__` as a name rather than the content.

ccordoba12: Nice!

spyder_kernels/console/tests/test_console_kernel.py
```diff
@@ -396,6 +396,21 @@ def test_runcell(tmpdir):
...
+        client.inspect('fname')
+        msg = client.get_shell_msg(block=True, timeout=TIMEOUT)
+        print(msg['content'])
+        content = msg['content']
+        assert "cell-test.py" in content['data']['text/plain']
+
+        # Verify that the `__file__` variable is undefined
+        client.inspect('__file__')
+        msg = client.get_shell_msg(block=True, timeout=TIMEOUT)
+        print(msg['content'])
```
ccordoba12: I think the `print`s in this test are unnecessary (the first one was my fault). Could you remove them to see what happens? Thanks!

bcolsen: It works fine without them. :+1:

PR #80, opened 2019-01-24
PR: Fix loading Spydata file with changed filename and other edge-cases in load_dict

CAM-Gerlach:
Before this PR, ``.spydata`` files could not be loaded if their filename was changed from the original value. This PR fixes that and allows such files to be loaded. Additionally, it cleans up a few smaller issues/edge cases in the ``load_dict()`` function:

* It now catches and reports ``AttributeErrors`` loading files with pickled functions that aren't currently loaded in the ``__main__`` namespace
* It always makes sure to clean up its temp directory and reset the working dir even if an uncaught error occurs 
* It uses a context manager for the tar file and ensures it is closed and cleaned up properly (it wasn't before), and
* It removes ancient pre-Spyder-2.2 code for loading Python 2 namespaces saved with the nearly 6 year old Spyder 2.0 and 2.1, that resulted in a ``DeprecationWarning``, slowed down the load process and could cause uncaught errors

Additionally, it adds tests/test coverage for renamed files, files with errors that prevent them being loaded, and that the working directory is reset properly.

Fixes #9

PR #81, opened 2019-01-24
PR: Fix and improve saving of Spyder namespace with many types of objects

### LINKED ISSUE #75: Namespace serialization silently fails if any object is unserializable, e.g. a Python module
A user reported on Gitter that Spyder's namespace saving feature was silently failing in certain situations, and upon testing it, an easily reproducible was was attempting to serialize Python modules, which cannot be serialized and are rarely if ever desired to be in this case. To fix this, we should:

* Fix error messages not being printed by handling exceptions in the ``deepcopy()`` call properly
* Call deepcopy on each individual value rather than the whole dict/namespace, to prevent falling back to normal ``copy`` or nothing at all for everything if one object cannot be deepcopied (or serialized entirely)
* Skip serializing anything that is a Python module EDIT: Or a function/class, as they can only be loaded if they already exist in the environment (thus it serves little purpose to do so, other than the prevent the rest of the environment from being loaded)
* If an object produces an error on attempted serialization, skip it and generate a popup warning listing the objects not serialized rather than refusing to save the entire session and forcing users to manually figure out and delete problem ones

@ccordoba12 These changes fix a major bug in namespace serialization, can all be implemented purely in Spyder-Kernels without any Spyder-side changes; and all in a file that has no changes (and thus no chance of merge conflicts) between ``0.x`` and ``master``, so can this go in ``0.x``?

CAM-Gerlach:
As a compliment to #80 on the loading namespaces side, this PR implements a number of fixes to make saving them work much more reliably and robustly, and improve the user experience when problems occur. While it is independent of any changes in Spyder, it is complimented by spyder-ide/spyder#8636 which makes the UI text less confusing should some objects fail to be serialized.

In particular, as proposed in #75 , this PR:

* Fixes error messages not being printed by handling exceptions in the ``deepcopy()`` call properly
* Calls deepcopy on each individual value rather than the whole dict/namespace, to prevent falling back to normal ``copy`` or nothing at all for everything if one object cannot be deepcopied (or serialized entirely)
* Skips serializing anything that is a Python module as they cannot be serialized, and skips callables (functions/classes), as they can only be loaded if they already exist in the environment (thus it serves little purpose to do so, other than the prevent the rest of the environment from being loaded)
* If an object produces an error on attempted serialization, skips it and generate a popup warning listing the objects not serialized rather than refusing to save the entire session and forcing users to manually figure out and delete problem ones
* Doesn't try to fall back to a normal copy() if deepcopy() fails, since this essentially means pickle() will also fail barring exceptional circumstances and could lead to the user's environment being modified without warning
* Saves future ``spydata`` files in POSIX.1-2001-standard ("Pax") Tar format for lesser restrictions on and more consistent/cross platform file/variable names and metadata, as well as better interoperability and standards compliance; uses a context manager to ensure the Tar file is always closed properly. This is transparent to users and Spyder; the loading process is identical and any supported version of Python supports Pax equally to GNU.
* Ensures the current working directory is always reset properly even if an unhandled exception occurs

Further, it adds a number of test fixtures, simplifies many of the text functions and makes the test of saving the namespace much more comprehensive with a number of test cases and additional checks.

Fixes #75

### DIFF
--- a/spyder_kernels/utils/iofuncs.py
+++ b/spyder_kernels/utils/iofuncs.py
@@ -22,6 +22,7 @@ import os.path as osp
 import tarfile
 import tempfile
 import shutil
+import types
 import warnings
 import json
 import inspect
@@ -289,14 +290,27 @@ def save_dictionary(data, filename):
     old_cwd = getcwd()
     os.chdir(osp.dirname(filename))
     error_message = None
+    skipped_keys = []
+    data_copy = {}
 
     try:
         # Copy dictionary before modifying it to fix #6689
-        try:
-            data = copy.deepcopy(data)
-        # Fall back to non-deep copying for dicts with objs not supporting it
-        except NotImplementedError:
-            data = copy.copy(data)
+        for obj_name, obj_value in data.items():
+            # Skip modules, since they can't be pickled, users virtually never
+            # would want them to be and so they don't show up in the skip list.
+            # Skip callables, since they are only pickled by reference and thus
+            # must already be present in the user's environment anyway.
+            if not (callable(obj_value) or isinstance(obj_value,
+                                                      types.ModuleType)):
+                # If an object cannot be deepcopied, then it cannot be pickled.
+                # Ergo, we skip it and list it later.
+                try:
+                    data_copy[obj_name] = copy.deepcopy(obj_value)
+                except Exception:
+                    skipped_keys.append(obj_name)
+        data = data_copy
+        if not data:
+            raise RuntimeError('No supported objects to save')
 
         saved_arrays = {}
         if load_array is not None:
@@ -338,6 +352,10 @@ def save_dictionary(data, filename):
             data.pop('__saved_arrays__')
     except (RuntimeError, pickle.PicklingError, TypeError) as error:
         error_message = to_text_string(error)
+    else:
+        if skipped_keys:
+            error_message = ('Some objects could not be saved: '
+                             + ', '.join(skipped_keys))
     finally:
         os.chdir(old_cwd)
     return error_message
--- a/spyder_kernels/utils/iofuncs.py
+++ b/spyder_kernels/utils/iofuncs.py
@@ -317,39 +317,64 @@ def save_dictionary(data, filename):
             # Saving numpy arrays with np.save
             arr_fname = osp.splitext(filename)[0]
             for name in list(data.keys()):
-                if isinstance(data[name], np.ndarray) and data[name].size > 0:
-                    # Saving arrays at data root
-                    fname = __save_array(data[name], arr_fname,
-                                       len(saved_arrays))
-                    saved_arrays[(name, None)] = osp.basename(fname)
-                    data.pop(name)
-                elif isinstance(data[name], (list, dict)):
-                    # Saving arrays nested in lists or dictionaries
-                    if isinstance(data[name], list):
-                        iterator = enumerate(data[name])
-                    else:
-                        iterator = iter(list(data[name].items()))
-                    to_remove = []
-                    for index, value in iterator:
-                        if isinstance(value, np.ndarray) and value.size > 0:
-                            fname = __save_array(value, arr_fname,
-                                               len(saved_arrays))
-                            saved_arrays[(name, index)] = osp.basename(fname)
-                            to_remove.append(index)
-                    for index in sorted(to_remove, reverse=True):
-                        data[name].pop(index)
+                try:
+                    if isinstance(data[name],
+                                  np.ndarray) and data[name].size > 0:
+                        # Save arrays at data root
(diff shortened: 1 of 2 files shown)


spyder_kernels/utils/iofuncs.py
```diff
@@ -288,58 +289,101 @@ def save_dictionary(data, filename):
...
 
-    # Copy dictionary before modifying to fix #6689
     try:
-        data = copy.deepcopy(data)
-    except NotImplementedError:
-        try:
-            data = copy.copy(data)
-        except Exception:
-            data = data
+        # Copy dictionary before modifying it to fix #6689
```
ccordoba12: If this PR fixes that Spyder issue, please update the description with the corresponding info.

The good thing about Github is that it can close issues among repos.

spyder_kernels/utils/iofuncs.py
```diff
@@ -288,58 +289,101 @@ def save_dictionary(data, filename):
...
+                                          np.ndarray) and value.size > 0:
+                                fname = __save_array(value, arr_fname,
+                                                     len(saved_arrays))
+                                saved_arrays[(name, index)] = (
+                                    osp.basename(fname))
+                                to_remove.append(index)
+                        for index in sorted(to_remove, reverse=True):
+                            data[name].pop(index)
+                    # If an array can't be saved with numpy for some reason,
+                    # leave the object intact and try to save it normally.
```
ccordoba12: Shouldn't this comment be above the `pass` statement below? I think the code would be easier to read that way.

spyder_kernels/utils/iofuncs.py
```diff
@@ -288,58 +289,101 @@ def save_dictionary(data, filename):
...
+                except (RuntimeError, pickle.PicklingError, TypeError):
+                    pass
             if saved_arrays:
                 data['__saved_arrays__'] = saved_arrays
-        pickle_filename = osp.splitext(filename)[0]+'.pickle'
+
+        pickle_filename = osp.splitext(filename)[0] + '.pickle'
+        # Attempt to pickle everything.
+        # If pickling fails, iterate through to eliminate problem objs & retry.
```
ccordoba12: `problem` -> `problematic`? (I don't know if that's the right word in this context or not).

spyder_kernels/utils/iofuncs.py
```diff
@@ -288,58 +289,101 @@ def save_dictionary(data, filename):
...
+                    except Exception:
+                        skipped_keys.append(obj_name)
+                    else:
+                        data_filtered[obj_name] = obj_value
+                if not data_filtered:
+                    raise RuntimeError('No supported objects to save')
+                pickle.dump(data_filtered, fdesc, protocol=2)
+
+        # Use PAX (POSIX.1-2001) format instead of default GNU.
+        # This improves interoperability and UTF-8/long variable name support.
```
ccordoba12: Is this change backwards compatible? I mean, what will happen with old spydata files? Will users be able to open them after this change?

spyder_kernels/utils/iofuncs.py
CAM-Gerlach: That comment was there before; it just got moved around with its accompanying code block. I already fixed that issue (although this PR implements a better actual fix). Thanks for checking.

spyder_kernels/utils/iofuncs.py
CAM-Gerlach: Yeah, you're right. I actually noticed it late last night but it was 5 am and I was just too tired to do another ``rebase`` to fix it. Fixed now.

spyder_kernels/utils/iofuncs.py
CAM-Gerlach: Either would work; "Problem <noun>" specifically refers to one or more items in a larger group of a specific type causing an issue, which seems appropriate here.

spyder_kernels/utils/iofuncs.py
CAM-Gerlach: Yup, 100%—I made sure of that. It has no effect on reading since any remotely modern version of the standard library ``tarfile`` module dating back to at least before Python 2.6 (which Spyder no longer supports anyway) can read PAX as well as the default GNU format, and the proper format to read it as detected automatically from the file when loading. It just means any new files saved will be a little less restrictive and more interoperable.

ccordoba12: Ok, good to know.

ccordoba12:
Ok, please fix the tests in PR #80 so I can merge it, then rebase this one to get the fix and we'll be good to go.

CAM-Gerlach:
@ccordoba12 ?

ccordoba12:
Didn't see this one, sorry. Merging.

CAM-Gerlach:
Thanks @ccordoba12 !

PR #92, opened 2019-03-03
PR: Include user site-packages directory in the list of excluded paths by the UMR

ccordoba12:
Fixes spyder-ide/spyder#8776


drumfiend2:
I posted this on issue #8776, but since that is an old and closed issue, I wanted to follow up on obbe79's question (https://github.com/spyder-ide/spyder/issues/8776) from that thread regarding Windows, as I'm currently having the same exact problem and was wondering if there was any resolution since this post. Thanks!

PR #98, opened 2019-03-23
PR: Add position to runcell

### LINKED ISSUE #97: Can't repeat runcell from terminal
When calling runfile, I can just repeat the call in the terminal by pressing up + enter.
I use to be able to do that when running a cell as well. But since the introduction of the runcell function, I get:
```
--Run Cell Error--
Please use only through Spyder's Editor; shouldn't be called manually from the console
```
Maybe it would make sense to give the line number to runcell so it could be called from the terminal?

impact27:
fixes  #97

### DIFF
--- a/spyder_kernels/customize/spydercustomize.py
+++ b/spyder_kernels/customize/spydercustomize.py
@@ -879,7 +879,8 @@ def runcell(cellname, filename, line_number=None):
                     position = it.start()
             if position == -1:
                 _print("--Run Cell Error--\n"
-                       f"Line {line_number} not found.")
+                       "Line {line_number} not found.".format(
+                               line_number=line_number))
                 return
             cell_boundary_re = re.compile(
                 r'(?:^\s*(?:# ?%%|# <codecell>|# In\[)|\A|\Z)',
--- a/spyder_kernels/customize/spydercustomize.py
+++ b/spyder_kernels/customize/spydercustomize.py
@@ -883,7 +883,7 @@ def runcell(cellname, filename, line_number=None):
                                line_number=line_number))
                 return
             cell_boundary_re = re.compile(
-                r'(?:^\s*(?:# ?%%|# <codecell>|# In\[)|\A|\Z)',
+                r'(?:^\s*(?:# ?%%|# <codecell>|# In\[.*\])|\A|\Z)',
                 re.MULTILINE)
             re_iter = cell_boundary_re.finditer(text)
             end_idx = -1
--- a/spyder_kernels/customize/spydercustomize.py
+++ b/spyder_kernels/customize/spydercustomize.py
@@ -836,7 +836,7 @@ def runfile(filename, args=None, wdir=None, namespace=None, post_mortem=False):
 builtins.runfile = runfile
 
 
-def runcell(cellname, filename):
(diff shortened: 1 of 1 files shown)


ccordoba12:
I'm a bit -1 about this because it makes no sense trying to run cells outside Spyder and in a different session. Cells are only a help for exploratory coding, nothing else.  

@bcolsen, what do you think about this one?

impact27:
My use case would mainly be to be able to repeat the previous cell with up + enter. 
I just feel that it is strange to have a function executed in the console, and added to the history, with hidden arguments, which means that you can’t execute it again.

impact27:
And also if I just executed a cell, there is no way for me to know which one.

impact27:
Also, that would give the option to run another cell from the console without having to click on the editor with the mouse. For example if I am looking at the code and want to run the cell at line 45.

bcolsen:
To start with, if you haven't already, please read through https://github.com/spyder-ide/spyder/issues/7113 to understand process that got us to the current runcell implementation. There it is mentioned that currently we can't have the qtconsole run commands in spyder without implementing the jupyter custom messaging api. This would be the best answer to this problem but it is quite difficult.

In this current solution you are relying on a saved file, which won't have the most up to date changes made to the file if the user hasn't saved. In fact since the runcell shortcut doesn't save the current file this solution won't necessarily even repeat the same code that was run with the last runcell shortcut. Even if the runcell shortcut saved the file running the runcell function in the console can't save the file in the editor. This is also a problem with the runfile function. A user might think that this function would save and run the file in the editor, but it doesn't.

I'm also -1 here.

> My use case would mainly be to be able to repeat the previous cell with up + enter.
I just feel that it is strange to have a function executed in the console, and added to the history, with hidden arguments, which means that you can’t execute it again.

This is my one documented regression with the runcell function. To be clear though, up + enter didn't work that well with the previous version. It would only rerun the same code that was copied in last time.

Having runcell repeat the last command was one option that was discussed but it would be ultimately more confusing for the users since they would think it was running the current version of the cell from there open code and not a version from last tie they press ctrl + enter

> And also if I just executed a cell, there is no way for me to know which one.

This is why the runcell function shows the cell name, but I can see that some people don't name their cells, so it's not a bad idea to add the line number into runcell for history purposes, but since line numbers in file can easily change users would be better off giving their cells meaningful unique names.

> Also, that would give the option to run another cell from the console without having to click on the editor with the mouse. For example if I am looking at the code and want to run the cell at line 45.

It would be cumbersome to type out the file name and as previously mentioned the line numbers would only match it the file was saved. The hot keys to go from the editor to the console are quite handy in this situation.

impact27:
I understand now why this is a bad solution. Maybe if the cell is unnamed, the title could be: "Unnamed cell @ line 37"?

impact27:
Title changed here: https://github.com/spyder-ide/spyder/pull/9014

ccordoba12:
So by naming cells this is not necessary anymore, right?

impact27:
It solves the history problem, but not the up + enter problem, but I can see now that this PR is not the correct approach for this problem.

ccordoba12:
Ok, thanks.

PR #101, opened 2019-04-01
Add line_number parameter

impact27:
Changes needed to make https://github.com/spyder-ide/spyder/pull/9014 work


impact27:
Not needed anymore

PR #103, opened 2019-04-04
PR: Enclose turtle customizations in a try/except to avoid a dependency on Tk

### LINKED ISSUE #102: Tkinter is now required for version 0.4.3 after patching the turtle code
With the merge of PR #90, Python now needs "to be configured for Tk" due to the explicit import of ```turtle``` (which in turn does ``` import Tkinter as TK```). I was updating ```py-spyder``` to version 3.3.4 in MacPorts and because of this I would need to add a dependency on ```py-tkinter```, which recursively pulls in ```tk```, ```tcl``` and their dependencies. We didn't do that before and whereas (admittedly) the same error would appear if someone selected the ```Tkinter``` backend, everything would work fine for the (default) "inline" or the "Qt5" backends.

I couldn't find the need of "having Python configured for Tk" in the requirements for spyder or spyder-kernels, did I miss this somewhere? I also suppose that the patching of the ```turtle``` code would only be relevant when choosing the Tkinter graphical backend and not for others - correct? If so, would it be possible to only do this when selecting the Tkinter backend? 

I'd appreciate your comments/suggestions and would welcome any advice you can give (regarding dependencies / backends) for downstream packagers.

ccordoba12:
Fixes #102.

### DIFF
--- a/spyder_kernels/customize/spydercustomize.py
+++ b/spyder_kernels/customize/spydercustomize.py
@@ -215,16 +215,19 @@ except:
 # This is needed to prevent turtle scripts crashes after multiple runs in the
 # same IPython Console instance.
 # See Spyder issue #6278
-import turtle
-from turtle import Screen, Terminator
+try:
+    import turtle
+    from turtle import Screen, Terminator
 
-def spyder_bye():
-    try:
-        Screen().bye()
-        turtle.TurtleScreen._RUNNING = True
-    except Terminator:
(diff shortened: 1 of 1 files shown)


reneeotten:
thanks for the quick response Carlos, that indeed fixes it!

ccordoba12:
No problem, thanks for testing. I'll release this new version during the weekend.

PR #104, opened 2019-04-06
PR: Handle option to show/hide cmd windows generated by the subprocess module

dalthviz:
Part of spyder-ide/spyder#9113

### DIFF
--- a/spyder_kernels/customize/spydercustomize.py
+++ b/spyder_kernels/customize/spydercustomize.py
@@ -44,7 +44,7 @@ if not hasattr(sys, 'argv'):
 # Main constants
 #==============================================================================
 IS_EXT_INTERPRETER = os.environ.get('SPY_EXTERNAL_INTERPRETER') == "True"
-SHOW_SUBPROCESS_WINDOW = os.environ.get('SPY_SHOW_CMD') == "True"
+HIDE_CMD_WINDOWS = os.environ.get('SPY_HIDE_CMD') == "True"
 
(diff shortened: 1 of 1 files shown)


spyder_kernels/customize/spydercustomize.py
```diff
@@ -44,7 +44,7 @@
 # Main constants
 #==============================================================================
 IS_EXT_INTERPRETER = os.environ.get('SPY_EXTERNAL_INTERPRETER') == "True"
-
+SHOW_SUBPROCESS_WINDOW = os.environ.get('SPY_SHOW_CMD') == "True"
```
ccordoba12: Let's change this constant name to `SHOW_CMD_WINDOWS`

spyder_kernels/customize/spydercustomize.py
```diff
@@ -139,7 +139,8 @@ def execfile(filename, namespace):
 
     class SubprocessPopen(subprocess.Popen):
         def __init__(self, *args, **kwargs):
-            kwargs['creationflags'] = creation_flag
+            if SHOW_SUBPROCESS_WINDOW:
```
ccordoba12: Let's move this to the top of the enclosing `if` block, like this:

```python
#==============================================================================
# Prevent subprocess.Popen calls to create visible console windows on Windows.
# See issue #4932
#==============================================================================
if os.name == 'nt' and SHOW_CMD_WINDOWS:
```

PR #106, opened 2019-04-07
PR: Skip test_turtle_launch if Tk is not installed

ccordoba12:


ccordoba12:
@sebix, is this what you need to avoid our tests to depend on Tk?

sebix:
This helps, yes. Thanks!

ccordoba12:
Ok, thanks for the confirmation.

PR #109, opened 2019-04-25
PR: Fix compiled module imports when starting kernels in a non-default interpreter

dalthviz:
Fixes spyder-ide/spyder#9077

### DIFF
--- a/spyder_kernels/console/start.py
+++ b/spyder_kernels/console/start.py
@@ -12,11 +12,10 @@ File used to start kernels for the IPython Console
 
 # Standard library imports
 from distutils.version import LooseVersion
-import locale
-import quote
+import json
 import os
 import os.path as osp
-import subprocess
+from subprocess import check_output
 import sys
 import site
 
@@ -87,33 +86,21 @@ def is_anaconda():
     return is_conda
 
 
-def exec_in_env(conda_root, envname, *command):
-    """
-    Run the standard conda activation script.
-    Taken from https://github.com/Anaconda-Platform/nb_conda_kernels/blob/
-    master/nb_conda_kernels/runner.py#L13
-    """
-    if sys.platform.startswith('win'):
-        print(sys.stdin.encoding, sys.stdout.encoding, sys.stderr.encoding,
-              locale.getpreferredencoding())
-        activate = os.path.join(conda_root, 'Scripts', 'activate.bat')
-        activator = subprocess.list2cmdline(['call', activate, envname])
-        ecomm = [os.environ['COMSPEC'], '/S', '/U', '/C', '@echo', 'off', '&&',
-                 'chcp', '&&', 'call', 'activate', envname,
-                 '&&'] + list(command)
-        subprocess.Popen(ecomm).wait()
-    else:
-        command = ' '.join(quote(c) for c in command)
-        activate = os.path.join(conda_root, 'bin', 'activate')
-        ecomm = ". '{}' '{}' && exec {}".format(activate, envname, command)
-        ecomm = ['sh' if 'bsd' in sys.platform else 'bash', '-c', ecomm]
-        os.execvp(ecomm[0], ecomm)
+def get_conda_root_path():
+    """Get conda root path using 'conda info --json'."""
+    conda_info = check_output('conda info --json')
+    conda_info_json = json.loads(conda_info)
+    return conda_info_json['env_vars']['CONDA_ROOT']
 
 
 def get_conda_env_path():
-    """Get conda environment PATH when running under a conda env."""
+    """
+    Get conda environment minimum PATH elements when running under a conda env.
+    """
     # TODO: Get the correct PATH that is setted when calling 'activate <env>'
     # by using exec_in_env or build it using sys.executable
+    env_root_path = sys.executable
+    conda_root_path = get_conda_root_path()
 
 
 def kernel_config():
@@ -335,9 +322,9 @@ def main():
         sys.path.remove('')
 
     # Handle conda environment. See Spyder issue #9077
-    if os.environ.get('SPY_EXTERNAL_INTERPRETER'):
+    if os.environ.get('SPY_EXTERNAL_INTERPRETER') and os.name == 'nt':
         if is_anaconda():
-            os.environ['PATH'] = get_conda_env_path()
+            os.environ['PATH'] = get_conda_env_path() + os.environ['PATH']
 
     # Fire up the kernel instance.
     from ipykernel.kernelapp import IPKernelApp
--- a/spyder_kernels/console/start.py
+++ b/spyder_kernels/console/start.py
@@ -19,6 +19,9 @@ from subprocess import check_output
 import sys
 import site
 
+# Local imports
+from spyder_kernels.py3compat import to_text_string
+
 
 PY2 = sys.version[0] == '2'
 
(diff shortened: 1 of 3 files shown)


spyder_kernels/console/start.py
```diff
@@ -87,33 +86,21 @@ def is_anaconda():
...
-        subprocess.Popen(ecomm).wait()
-    else:
-        command = ' '.join(quote(c) for c in command)
-        activate = os.path.join(conda_root, 'bin', 'activate')
-        ecomm = ". '{}' '{}' && exec {}".format(activate, envname, command)
-        ecomm = ['sh' if 'bsd' in sys.platform else 'bash', '-c', ecomm]
-        os.execvp(ecomm[0], ecomm)
+def get_conda_root_path():
+    """Get conda root path using 'conda info --json'."""
+    conda_info = check_output('conda info --json')
```
goanpeca: This can fail so we need to catch any possible errors.

You can test this by using any random thing as command

`condanothere info --json`

spyder_kernels/console/start.py
```diff
@@ -87,33 +86,21 @@ def is_anaconda():
...
+    conda_info_json = json.loads(conda_info)
```
goanpeca: This could also potentially fail so we need to cathc this as well

spyder_kernels/console/start.py
```diff
@@ -87,33 +86,21 @@ def is_anaconda():
...
+    return conda_info_json['env_vars']['CONDA_ROOT']
```
goanpeca: Just is case is better to also use `dict.get` and add ifs to act in a more safe maner

spyder_kernels/console/start.py
```diff
@@ -75,6 +77,53 @@ def sympy_config(mpl_backend):
...
+        return conda_info_json['env_vars']['CONDA_ROOT']
+    except Exception:
+        return None
+
+
+def get_conda_env_path():
+    """
+    Get conda environment minimum PATH elements when running under a conda env.
+    """
+    env_root_exec = sys.executable
```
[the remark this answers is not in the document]
ccordoba12: I think's also better to apply `to_text_string` to `sys.executable`, use `u'Library'`, etc, anywhere, and then transform the final result to a binary string for Python 2.

That way we will avoid any problems with environments created in directories with non-ascii characters.

spyder_kernels/console/start.py
```diff
@@ -75,6 +77,53 @@ def sympy_config(mpl_backend):
...
+        conda_info = check_output('conda info --json')
+        conda_info_json = json.loads(conda_info)
+        return conda_info_json['env_vars']['CONDA_ROOT']
+    except Exception:
+        return None
+
+
+def get_conda_env_path():
+    """
+    Get conda environment minimum PATH elements when running under a conda env.
```
ccordoba12: This should be

    Get minimal set of directories to add to PATH for conda environments on Windows.

spyder_kernels/console/start.py
```diff
@@ -75,6 +77,53 @@ def sympy_config(mpl_backend):
...
+def is_anaconda():
+    """
+    Detect if we are running under Anaconda.
+    Taken from https://stackoverflow.com/a/47610844/438386
+    """
+    is_conda = osp.exists(osp.join(sys.prefix, 'conda-meta'))
+    return is_conda
+
+
+def get_conda_root_path():
```
ccordoba12: Let's change this to `get_conda_rootpath`

ccordoba12:
I forgot to mention two things:

1. Please rebase against 0.x to get the fix for test that's failing.
2. Please see this SO post about how to set environment variables for Python 2 and 3:
    https://stackoverflow.com/a/31003672/438386

spyder_kernels/console/start.py
```diff
@@ -75,6 +80,56 @@ def sympy_config(mpl_backend):
     return lines
 
 
+def is_anaconda():
```
ccordoba12: @dalthviz, please move this function and the other functions you added below to `utils/anaconda.py` and import the necessary ones from there.

spyder_kernels/console/start.py
```diff
@@ -293,6 +348,17 @@ def main():
     while '' in sys.path:
         sys.path.remove('')
 
+    # Handle conda environment. See Spyder issue #9077
```
ccordoba12: Let's change this comment to

    # Add to PATH most of the directories added when activating a conda
    # environment.
    # See Spyder issue #9077.

goanpeca: @ccordoba12 With URL awareness work soon in place, it would really help when browsing the code. How about adding the full link?

```python
# Add to PATH most of the directories added when activating a conda
# environment.
# See https://github.com/spyder-ide/spyder/issues/9077
```

ccordoba12:
@CAM-Gerlach, please help us test this PR to see if it fixes spyder-ide/spyder#9077, which I think is the same problem you reported some time ago about our kernel not working for Python compiled packages when an external interpreter is selected.

Thanks!

CAM-Gerlach:
@ccordoba12 Yup, this is almost certainly the same problem, and the fix is essentially what I would expect would fix it.

One thing: In this context, does "external interpreters" mean kernels launched by Spyder, but in an environment different from Spyder's own? I'd always been using it to mean kernels that were launched outside (externally) of Spyder, and then connected to it, regardless of the environment. It is the former that both I and the OP are referring to; doing the latter *does* work and is in fact the workaround I was using to this problem. Therefore, this fix should apply to the former case, but not the latter. Is that the case?

 I'll be free to test it an a couple hours, thanks.

spyder_kernels/utils/anaconda.py
```diff
@@ -0,0 +1,69 @@
+# -*- coding: utf-8 -*-
+# -----------------------------------------------------------------------------
+# Copyright (c) 2009- Spyder Kernels Contributors
+#
+# Licensed under the terms of the MIT License
+# (see spyder_kernels/__init__.py for details)
+# -----------------------------------------------------------------------------
+
+"""Utilities for conda environtment handling"""
```
CAM-Gerlach: Period

spyder_kernels/utils/anaconda.py
```diff
@@ -0,0 +1,69 @@
...
+import sys
+
+# Local library imports
+from spyder_kernels.py3compat import to_text_string
+
+
+def is_anaconda():
+    """
+    Detect if we are running under Anaconda.
+    Taken from https://stackoverflow.com/a/47610844/438386
```
CAM-Gerlach: Needs a blank line between this and the above.

spyder_kernels/utils/anaconda.py
```diff
@@ -0,0 +1,69 @@
...
+        conda_info_json = json.loads(conda_info)
+        return to_text_string(conda_info_json['env_vars']['CONDA_ROOT'])
+    except Exception:
+        return None
+
+
+def get_win_conda_envpath():
+    """
+    Get minimal set of directories to add to PATH for conda
+    environments on Windows.
```
CAM-Gerlach: ```Get directories to add to PATH for conda envs on Windows.``

ccordoba12:
> In this context, does "external interpreters" mean kernels launched by Spyder, but in an environment different from Spyder's own?

I modify the PR title a little bit, but yeah, that's what we're trying to fix here.

> but the issue is unfortunately not fixed as I get the same error in this scenario

I thought adding the env PATH to our kernel would be enough to fix this. Please post on the issue referenced above the error you're getting when importing compiled modules. Thanks!

CAM-Gerlach:
> I thought adding the env PATH to our kernel would be enough to fix this

Yeah, so did I, but I did observe before if I hacked the paths on to ``os.environ['PATH']`` (or ``sys.path``) manually in the console, it wouldn't work. Presumably, the DLL search path is not necessarily = ``PATH`` in this case, or it is set in stone before/during the launch of the interpreter itself. I'm guessing it might be something earlier/lower-level that needs to be set (e.g. with ``conda activate``), or the DLL search order needs to be updated directly.

@mingwandroid or @msarahan , any insight here?

> Please post on the issue referenced above the error you're getting when importing compiled modules.

I just get the same error as the OP with Numpy, or with a different module name with stuff like ``NetCDF4``, ``Cartopy`` and ``PyART``, or anything based on the PROJ4/GDAL/GEOS stack (GeoPandas, Fiona, Shaply, etc). Is there something more specific you wanted me to post there? Thanks.

CAM-Gerlach:
> does "external interpreters" mean kernels launched by Spyder, but in an environment different from Spyder's own?

@ccordoba12 This contradicts the usage in our own [docs](https://docs.spyder-ide.org/ipythonconsole.html#connect-to-an-external-kernel) and in our UI in the "Connect to existing kernels" dialog, using "external kernels/interpreters" to mean, as would linguistically make sense, kernels launched "externally" to Spyder (which in fact, work just fine). Therefore, I will update the PR and issue title to be more clear.

dhirschfeld:
Just updated to the latest spyder and ran into this problem.

Are there currently any blockers to merging this PR?

ccordoba12:
This is not the right solution, as @CAM-Gerlach pointed out. We really need to activate the conda env where the kernel is installed before starting it.

That's top in my priorities for spyder-kernels 1.9

PR #110, opened 2019-04-25
PR: Set Matplotlib backend to inline for kernels started outside Spyder

### LINKED ISSUE #108: Set matplotlib backend to inline by default on starting a new kernel
Opened by request of @ccordoba12 

``spyder-kernels`` should set the graphics backend to ``inline`` by default on starting a ``spyder-kernels`` kernel, to match the Spyder default and so plots work by default on remote machines over ``ssh``. Currently, in my testing, plots in such consoles appear in interactive mode (i.e. Qt backend), rather than inline.

ccordoba12:
Fixes #108.

### DIFF
--- a/.circleci/install.sh
+++ b/.circleci/install.sh
@@ -2,7 +2,7 @@
 
 export TRAVIS_OS_NAME="linux"
 export CONDA_DEPENDENCIES_FLAGS="--quiet"
-export CONDA_DEPENDENCIES="ipykernel cloudpickle nomkl numpy pandas scipy pytest pytest-cov mock cython"
+export CONDA_DEPENDENCIES="ipykernel cloudpickle nomkl numpy pandas scipy pytest pytest-cov mock cython matplotlib"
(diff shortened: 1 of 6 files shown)


ccordoba12:
> Its in Spyder 4 with external kernels where this matters, as the backend currently needs to be explicitly set to inline

Ok, I was not aware of that. I'll move it to master then.

CAM-Gerlach:
Thanks; I should have been clearer and more explicit about that in my initial issue. I was in a rush to go somewhere so I wasn't as thorough as usual. I'll review this again.

ccordoba12:
Ok, thanks for the confirmation. I still don't understand how things are working correctly for Spyder 3 and not for 4, so I'm going to move back this PR to `0.x` to be totally sure.

CAM-Gerlach:
> I still don't understand how things are working correctly for Spyder 3 and not for 4

It has something to do with the Plots plugin and not displaying figures inline due to that, as far as I'm aware.

PR #111, opened 2019-05-02
PR: Use Jupyter comms to communicate with the Spyder frontend

impact27:
API to communicate between the spyder IDE and the spyder kernel.
It uses IPython Comms for messaging. The messages are sent by calling an
arbitrary function, with the limitation that the arguments have to be
picklable. If the function must return, the call must be blocking.

In addition, the frontend can interrupt the kernel to process the message sent.
This allows, for example, to set a breakpoint in pdb while the debugger is
running. The signal signal.SIGUSR1 is used. Signaling is not supported on
Windows, so the message will only be delivered when the kernel is checking the
event loop, or if pdb is waiting for an input.

Example:

On one side:

    ```
    def hello_str(msg):
        print('Hello ' + msg + '!')

    def add(a, d):
        return a + b

    left_comm.register_call_handler('add_numbers', add)
    left_comm.register_call_handler('print_hello', hello_str)
    ```

On the other:

    ```
    right_comm.remote_call().print_hello('world')
    res =  right_comm.remote_call(blocking=True).add_numbers(1, 2)
    print('1 + 2 = ' + str(res))
    ```

Which prints on the right side (The one with the `left_comm`):

    ```
    Hello world!
    ```

And on the left side:

    ```
    1 + 2 = 3
    ```

This may be a problem if the user is using `SIGUSR1`. Should we have an option to :
a) Change the signal used?
b) disable the signaling?


impact27:
For reference: https://ipython.org/ipython-doc/stable/development/messaging.html

impact27:
I don't understand why the matplotlib tests are failing

ccordoba12:
Please merge with master to get those tests fixed.

impact27:
I think this is ready for review. The test is failing because 
`ModuleNotFoundError: No module named 'wurlitzer' `

impact27:
Are these tests running on windows? If that is the case I might need some help.

impact27:
There must be something I am missing about how signals are handled in windows. `spyder_kernels/comms/tests/test_comms.py::test_signal` is failing. Does this work on windows?
```python
def signal_handler(signum, frame):
    pass
signal.signal(signal.SIGFPE, signal_handler)
os.kill(os.getpid(), signal.SIGFPE)
```

impact27:
I will just not support the interruptions for windows for now.

impact27:
For reference: https://stackoverflow.com/questions/35772001/how-to-handle-the-signal-in-python-on-windows-machine

impact27:
The tests are finally passing, so this is ready for review.

spyder_kernels/comms/__init__.py
```diff
@@ -0,0 +1,11 @@
+# -*- coding: utf-8 -*-
+# -----------------------------------------------------------------------------
+# Copyright (c) 2009- Spyder Kernels Contributors
+#
+# Licensed under the terms of the MIT License
+# (see spyder_kernels/__init__.py for details)
+# -----------------------------------------------------------------------------
+
+"""
+Comms
```
goanpeca:
```suggestion
Let's add something like:

```
Comms allow custom messages between the frontend and the kernel.


They are used, for instance, in ipywidgets to update widget state.

A comm consists of a pair of objects, in the kernel and the frontend, with an automatically assigned unique ID. When one side sends a message, a callback on the other side is triggered with that message data. Either side, the frontend or kernel, can open or close the comm.

See: https://jupyter-notebook.readthedocs.io/en/stable/comms.html
```

Comment (impact27) [reply spyder_kernels/comms/__init__.py]: Well what I call comms here are wrappers around the jupyter comms, not the comms themselves. the jupyter comms are stored in CommsBase._comm

Comment (goanpeca) [reply spyder_kernels/comms/__init__.py]: Hmm ok, but let's add something to that docstring. Feels a bit bare bones :-p

Comment (ccordoba12) [reply spyder_kernels/comms/__init__.py]: @impact27, you can just add what you put in your comment above. That would be fine.

**Comment by impact27** [comment]
There is a problem setting the breakpoints when signalling is disabled. I will fix that.

**Comment by impact27** [comment]
This should work now.

**Comment by impact27** [comment]
What I did is: if signalling is disabled (eg Windows) and the debugger is running, I use the debugger to process the queued messages. This allow the breakpoints to be correctly set.

**Comment by impact27** [comment]
This PR is changing the way messages are sent between the frontend and the kernel. The changes are implemented in https://github.com/spyder-ide/spyder/pull/9343. More specifically:

 -  `spyder/plugins/ipythonconsole/widgets/debugging.py`: `DebuggingWidget`'s `set_spyder_breakpoints` is using this API to send the breakpoints

 - `spyder/plugins/ipythonconsole/widgets/namespacebrowser.py` : `NamepaceBrowserWidget`'s `_handle_spyder_msg` is using this api to handle `'data'`, `'pdb_state'`, `'pdb_continue'`, and `'set_breakpoints'` messages. 

These changes were made to avoid duplication of the pickle code. This also mean that the current version of the kernel is not compatible with spyder pre-#9343, and #9343 is not compatible with the current master.

The question I have is therefore:
 - Should I make a backward incompatible change? Spyder4 is still in beta after all.
 - Should I duplicate the code so that this kernel can be used with a previous version of spyder? For spyder, I can just change the requirements for spyder-kernels. Should spyder-kernels also have requirments for spyder? Or equivalently should they come packaged together?

File: spyder_kernels/console/kernel.py [hunk 2]
```diff
@@ -118,47 +125,18 @@ def get_var_properties(self):
...
-        )
-        self.log.debug(msg)
-
     def get_value(self, name):
         """Get the value of a variable"""
         ns = self._get_current_namespace()
         value = ns[name]
         try:
-            self.send_spyder_msg('data', data=value)
+            self.frontend_comm.send_message('data', data=value)
```
impact27: should I use the "remote_call" mechanism instead of the "send_message" one?

spyder_kernels/comms/commbase.py
```diff
@@ -0,0 +1,344 @@
...
+            A function to handle the request.
+            Pass None to unregister the call_name
+        """
+        if not handler:
+            self._remote_call_handlers.pop(call_name, None)
+            return
+
+        self._remote_call_handlers[call_name] = handler
+
+    def send_message(self, spyder_msg_type, content=None, data=None):
```
impact27: The `send_message` is almost a duplicate of the `remote_call` mechanism. As "There should be only one obvious way to do it", should `send_message` and `register_message_handler` be made private?

spyder_kernels/comms/commbase.py
```diff
@@ -0,0 +1,344 @@
...
+        spyder_msg_type: str
+            The spyder message type
+        content: dict
+            The (JSONable) content of the message
+        data: any
+            Any object that is serializable by cloudpickle (should be most
+            things). Will arrive as cloudpickled bytes in `.buffers[0]`.
+        """
+        if not self.is_open():
+            raise RuntimeError("The comm is not connected.")
```
impact27: Should this be a warning?

spyder_kernels/comms/commbase.py
```diff
@@ -0,0 +1,344 @@
...
+        except Exception as e:
+            self._set_call_return_value(buffer, e, exception=True)
+
+    def _remote_callback(self, call_name, call_args, call_kwargs):
+        """Call the callback function for the remote call."""
+        if call_name in self._remote_call_handlers:
+            return self._remote_call_handlers[call_name](
+                *call_args, **call_kwargs)
+
+        raise RuntimeError("No such spyder call type: %s" % call_name)
```
impact27: Should this be a warning?

spyder_kernels/customize/spydercustomize.py
```diff
@@ -783,13 +793,25 @@ def _get_globals():
...
+
+
 def runfile(filename, args=None, wdir=None, namespace=None, post_mortem=False):
     """
     Run filename
     args: command line arguments (string)
     wdir: working directory
     post_mortem: boolean, whether to enter post-mortem mode on error
     """
+    if _comms_enabled():
```
impact27: This test is made for backward compatibility reasons. It can be removed if it is assumed that the comm must be open. Alternatively, The check could always be made for blocking calls, and non-blocking calls could be warnings.

spyder_kernels/comms/commbase.py
impact27: This is triggered if an unregistered function is called. It could be that only blocking calls should be an error, and otherwise this is only a warning?

spyder_kernels/console/kernel.py
impact27: I will do that

spyder_kernels/comms/commbase.py
impact27: I will do that as well

impact27:
I removed the unix interruptions from this PR, so I think this is pretty much done.

impact27:
I think that is everything, you can giuve it another look. All the tests for comms are now in spyder.

ccordoba12:
@impact27, please stop force-pushing in this PR because it's not possible for me to understand what has changed between commits.

Also, I'll continue my review here instead of in PR #112.

ccordoba12:
@impact27, please **stop** force-pushing. You're making almost impossible for me to review your work by doing that.

ccordoba12:
I mean, instead of force-pushing, please continue adding more commits here. If that breaks your other PRs built of top of this one, then you'll have to cherry-pick your commits in them or manually create new PRs with the code you added in them.

Sorry for the inconvenience, but that's the only way forward for me to review and merge the rest of your work.

spyder_kernels/comms/commbase.py
```diff
@@ -0,0 +1,372 @@
...
+                msg_dict, buffer, load_exception)
+            return
+
+        logger.debug("No such spyder message type: %s" % spyder_msg_type)
+
+    def _handle_remote_call(self, msg, buffer, load_exception):
+        """Handle a remote call."""
+        msg_dict = msg['content']
+        if load_exception:
+            raise load_exception
```
ccordoba12: From #112:

> If something doesn't happen, it can be frustrating for the user. As a user, I would rather know that something went wrong and be able to report it rather than not understanding why something is not happening.

I understand that, but then we should display the exception when Spyder is running in debug mode (so we'd to inform the kernel about that). Or reroute these exceptions to Spyder and show them through our error report dialog.

Otherwise I doubt people (other than Spyder developers, I mean) would be able to make anything out of these exceptions.

spyder_kernels/console/tests/test_console_kernel.py
```diff
@@ -22,9 +23,9 @@
 
 # Local imports
 from spyder_kernels.py3compat import PY3, to_text_string
-from spyder_kernels.console.kernel import PICKLE_PROTOCOL
+from spyder_kernels.comms.commbase import PICKLE_PROTOCOL
 from spyder_kernels.utils.iofuncs import iofunctions
-from spyder_kernels.utils.test_utils import get_kernel, get_log_text
+from spyder_kernels.utils.test_utils import get_kernel, get_log_text, dummyComm
```
ccordoba12: `dummyComm` is no longer used here. Would it be better to move this to Spyder?

spyder_kernels/customize/spydercustomize.py
```diff
@@ -303,16 +308,34 @@ def _patched_preparation_data(name):
...
         """Ask Spyder for breakpoints before the first prompt is created."""
-        if self.starting:
-            get_ipython().kernel._ask_spyder_for_breakpoints()
+        try:
+            _frontend_request(blocking=False).set_debug_state(True)
+            if self.starting:
+                breakpoints = _frontend_request().get_breakpoints()
+                self.set_spyder_breakpoints(breakpoints)
+        except (CommError, TimeoutError):
+            logger.debug("Could not get breakpoint from the frontend.")
```
ccordoba12: Sorry, please change here `breakpoint` for `breakpoints` (it was my mistake in #112).

spyder_kernels/customize/spydercustomize.py
```diff
@@ -303,16 +308,34 @@ def _patched_preparation_data(name):
...
+
+    def postloop(self):
+        """Notifies spyder that the loop has ended."""
+        try:
+            _frontend_request(blocking=False).set_debug_state(False)
+        except (CommError, TimeoutError):
+            logger.debug("Could not send postloop to frontend.")
```
ccordoba12: Please change this to

    Could not send debugging state to the frontend.

----

From #112:

> Aldo this could be useful in spyder to make decisions on debugging. Now in the spyder code, _reading is taken as meaning _reading_pdb in many places, but you could have a call to input() where it might not be true. It wold therefore be useful to have a way of telling if debugging is really taking place.

Yes, I totally agree that using `_reading` is quite brittle because a call to `input()` also sets that attribute to True. So it's really nice that we're finally able to differentiate between `_reading` and debugging proper. Thanks a lot for that!

Are you using this in other places in PR spyder-ide/spyder#9343? (I didn't see that).

spyder_kernels/customize/spydercustomize.py
```diff
@@ -783,13 +818,25 @@ def _get_globals():
...
     Run filename
     args: command line arguments (string)
     wdir: working directory
     post_mortem: boolean, whether to enter post-mortem mode on error
     """
+    try:
+        # Save the open files
+        _frontend_request().savefiles()
+    except (CommError, TimeoutError):
+        logger.debug("Could not send save files before executing.")
```
ccordoba12: I don't think this should be part of this PR. It's new functionality, so it should go in a different PR.

spyder_kernels/utils/test_utils.py
```diff
@@ -45,3 +45,26 @@ def get_kernel(kernel_class=SpyderKernel):
 def get_log_text(kernel):
     """Get the log of the given kernel."""
     return kernel.log.handlers[0].stream.getvalue()
+
+
+class dummyComm():
```
ccordoba12: As I said above, I think this should be moved to Spyder now.

impact27:
Sorry about pushing the PR. Should I instead add commits on top of the PR? I thought you wanted me to git-rebase to keep the history clean.

impact27:
> I mean, instead of force-pushing, please continue adding more commits here. If that breaks your other PRs built of top of this one, then you'll have to cherry-pick your commits in them or manually create new PRs with the code you added in them.
> 
> Sorry for the inconvenience, but that's the only way forward for me to review and merge the rest of your work.

Ok I will do that.

ccordoba12:
> Should I instead add commits on top of the PR?

Yes, please do.

> I thought you wanted me to git-rebase to keep the history clean.

Well, there's no need of that as long as your commit messages are more or less informative. I recall now that we suggested you in the past to squash most of your commits, but that's because your messages were not meaningful.

But anyway, even if your messages are not so descriptive, please keep adding more commits here so I understand how your work is evolving. Thanks!

ccordoba12:
> Ok I will do that.

Thanks a lot for understanding!

I think you should be able to sort things in your other PRs by doing a rebase on top of master after this PR is merged. If not, cherry-pick will certainly be your friend.

spyder_kernels/customize/spydercustomize.py
impact27: No

ccordoba12: Ok, no problem. We can add another handler later for Spyder to query the kernel if it's in debugging or not.

requirements/posix.txt
```diff
@@ -1,3 +1,3 @@
 ipykernel
 cloudpickle
-wurlitzer
\ No newline at end of file
+wurlitzer>=1.0.3
```
ccordoba12: Please add this requirement to our [setup.py](https://github.com/spyder-ide/spyder-kernels/blob/master/setup.py#L42). It's very important to have it there too.

impact27:
Is the coverage failure a problem? I moved the comms test to spyder but I can probably move some back here.

ccordoba12:
> Is the coverage failure a problem?

No, don't worry about them.

spyder_kernels/comms/frontendcomm.py
```diff
@@ -0,0 +1,71 @@
...
+        """Wait until the frontend replies to a request."""
+        if call_id in self._call_reply_dict:
+            return
+
+        # There is no get_ident in Py2
+        if not PY2 and self._main_thread_id != threading.get_ident():
+            # We can't call kernel.do_one_iteration from this thread.
+            # And we have no reason to think the main thread is not busy.
+            raise CommError(
+                "Can't make blocking calls from non main threads.")
```
ccordoba12: non main -> non-main

spyder_kernels/comms/commbase.py
```diff
@@ -0,0 +1,407 @@
...
+            The spyder message type
+        content: dict
+            The (JSONable) content of the message
+        data: any
+            Any object that is serializable by cloudpickle (should be most
+            things). Will arrive as cloudpickled bytes in `.buffers[0]`.
+        comm_id: int
+            the comm to send to. If None sends to all comms.
+        """
+
```
ccordoba12: Please remove this blank line.

spyder_kernels/comms/commbase.py
```diff
@@ -0,0 +1,407 @@
...
+                'call_name': The function name (mostly for debugging)
+                }
+"""
+
+import cloudpickle
+import logging
+import sys
+import uuid
+import traceback
+
```
ccordoba12: Please add another blank line below this one.

spyder_kernels/comms/commbase.py
```diff
@@ -0,0 +1,407 @@
...
+PY2 = sys.version[0] == '2'
+logger = logging.getLogger(__name__)
```
ccordoba12: Please add a blank line below this one.

spyder_kernels/comms/commbase.py
```diff
@@ -0,0 +1,407 @@
...
+
+        # Get message dict
+        msg_dict = msg['content']['data']
+
+        spyder_msg_type = msg_dict['spyder_msg_type']
+
+        if spyder_msg_type in self._message_handlers:
+            self._message_handlers[spyder_msg_type](
+                msg_dict, buffer, load_exception)
+            return
```
ccordoba12: Let's remove this `return` here and instead add an `else` in this line, like this:

```python
else:
    logger.debug("No such spyder message type: %s" % spyder_msg_type)
```

spyder_kernels/comms/commbase.py
```diff
@@ -0,0 +1,407 @@
...
+        try:
+            return_value = self._remote_callback(
+                    msg_dict['call_name'],
+                    buffer['call_args'],
+                    buffer['call_kwargs'])
+            self._set_call_return_value(msg_dict, return_value)
+        except Exception as e:
+            tb = traceback.extract_tb(sys.exc_info()[2])
+            self._set_call_return_value(msg_dict, e, traceback=tb)
```
ccordoba12: Are you doing something with the traceback on the Spyder side?

spyder_kernels/comms/commbase.py
```diff
@@ -0,0 +1,407 @@
...
+
+    def _remote_callback(self, call_name, call_args, call_kwargs):
+        """Call the callback function for the remote call."""
+        if call_name in self._remote_call_handlers:
+            return self._remote_call_handlers[call_name](
+                *call_args, **call_kwargs)
+
+        raise CommError("No such spyder call type: %s" % call_name)
```
ccordoba12: Let's change this to

```python
else:
    logger.debug("No such spyder call type: %s" % call_name)
```

as above.

spyder_kernels/comms/commbase.py
```diff
@@ -0,0 +1,407 @@
...
+
+        self._wait_reply(call_id, call_name, timeout)
+
+        reply = self._call_reply_dict[call_id]
+        self._call_reply_dict.pop(call_id)
+
+        if reply['is_error']:
+            error, tb = reply['value']
+            traceback.print_list(tb)
+            raise error
```
ccordoba12: Do we need to raise an error here? Can't we log it instead?

spyder_kernels/comms/commbase.py
```diff
@@ -0,0 +1,407 @@
...
+            handle_error = True
+        else:
+            handle_error = content['is_error']
+
+        self._call_reply_dict[request_id] = {
+                'is_error': handle_error,
+                'value': buffer,
+                'content': content
+                }
+        self._reply_recieved(request_id)
```
ccordoba12: _reply_recieved -> _reply_received

spyder_kernels/comms/frontendcomm.py
```diff
@@ -0,0 +1,71 @@
...
+        t_start = time.time()
+        self._wait_list[call_id] = call_name
+        while call_id not in self._call_reply_dict:
+            if time.time() > t_start + timeout:
+                raise TimeoutError(
```
ccordoba12: Could we log this instead of raising an error?

spyder_kernels/console/kernel.py
```diff
@@ -114,71 +166,25 @@ def get_var_properties(self):
...
-
-        # We send serialized values in a list of one element
-        # from Spyder to the kernel, to be able to send them
-        # at all in Python 2
-        svalue = value[0]
-
-        # We need to convert svalue to bytes if the frontend
-        # runs in Python 2 and the kernel runs in Python 3
-        if PY2_frontend and not PY2:
-            svalue = bytes(svalue, 'latin-1')
```
ccordoba12: How are you handling this case where the frontend runs in Python 2 and the kernel in Python 3?

It seems that you simply dismissed what we already had done before.

spyder_kernels/customize/spydercustomize.py
```diff
@@ -23,10 +23,15 @@
...
 import warnings
+import logging
 
 from IPython.core.getipython import get_ipython
 
+from spyder_kernels.py3compat import TimeoutError
+from spyder_kernels.comms import CommError
 
+
+logger = logging.getLogger(__name__)
```
ccordoba12: Please add a blank line below this one.

spyder_kernels/customize/spydercustomize.py
```diff
@@ -783,13 +818,22 @@ def _get_globals():
     return ipython_shell.user_ns
 
 
+def _frontend_request(blocking=True):
```
ccordoba12: Please add a docstring to this function.

spyder_kernels/comms/commbase.py
impact27: This is COmmbase so it applies for both sides. The idea is to send it to the other side to be handled.

spyder_kernels/comms/commbase.py
impact27: I want to notify the other side and the exception is catched above.

spyder_kernels/comms/commbase.py
impact27: So if the frontend calls the kernel, and an exception happens in the kernel, the exception is snet back to the frontend to be handled.

spyder_kernels/comms/commbase.py
impact27: Well maybe the calling function want to be able to catch that exception and do something with it. For example, get_value would raise a KeyError if the variable is not defines.

spyder_kernels/comms/commbase.py
ccordoba12: Sure, I understand that. For now, I think the only exceptions we need to show in the frontend are those generated when calling `get_value` (as I mentioned in PR spyder-ide/spyder#9343). The others are not so important for now.

spyder_kernels/console/kernel.py
impact27: Well the value could be anything, not just a string. the pickling is done in the comms. This function doesn't work with string anymore.

spyder_kernels/comms/frontendcomm.py
impact27: That is the same problem than in the other side: What is the return value of a blocking call if the request times out? I think this should be handled by the calling code.

spyder_kernels/console/kernel.py
impact27: This might be something that needs to be fixed in the comms instead.

impact27: And as is it doesn't work for me as the encoding I get from cloudpickle.dumps in python 2 is `utf8` and not `latin-1`.

spyder_kernels/comms/commbase.py
impact27: Ok but should I then add a try: except: in the other calls? I think it makes more sense than not raising this exception.

spyder_kernels/console/kernel.py
ccordoba12: This was not added here to handle strings. `svalue` was the serialized value of an object sent by Spyder to be set here in the kernel, after making changes to it through one of our editors (e.g. the dataframe editor).

The problem we solved here is that if the frontend is running in Python 2 and the kernel in Python 3, we need to convert that serialized value to `latin-1` before trying to deserialize it. Else we won't get the right value.

Now, since you move all communications to use serialization (before it was a mix of serialization and passing strings), I think you need to do this to before deserializing anything that comes from frontend and fulfills this case.

And you should also apply this trick in the frontend, if that runs in Python 3 and the kernel in Python 2 (but that could be done in CommsBase, right?)

spyder_kernels/comms/commbase.py
ccordoba12: That example makes no sense to me. `get_value` is triggered here by doing a double-click on the Variable Explorer. So if a KeyError is generated here, it means that we're not even displaying the current variables in the kernel namespace correctly.

However, the most important question is: what would users gain by seeing this kind of errors?

spyder_kernels/comms/frontendcomm.py
ccordoba12: I disagree. I think we should simplify things and make this return None.

What could the calling code do with a TimeoutError?

spyder_kernels/comms/commbase.py
ccordoba12: But that needs to be done in the frontend side, right?

spyder_kernels/console/kernel.py
impact27: What I mean is that there might be other problems in encoding:

In python 2:
```python
cloudpickle.dumps("eéè", 2)
'\x80\x02U\x05e\xc3\xa9\xc3\xa8q\x00.'
```

In python3:
```python
value = '\x80\x02U\x05e\xc3\xa9\xc3\xa8q\x00.'
bvalue = bytes(value, 'latin-1')
cloudpickle.loads(bvalue)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 1: ordinal not in range(128)
cloudpickle.loads(bvalue, encoding='latin-1')
'eÃ©Ã¨'
cloudpickle.loads(bvalue, encoding='utf8')
'eéè'
```

impact27: from https://docs.python.org/3/library/pickle.html#pickle.loads :

> 
    Optional keyword arguments are fix_imports, encoding and errors,
    which are used to control compatibility support for pickle stream 
    generated by Python 2.

ccordoba12: Ok, then this means we need to use the same encoding to serialize and deserialize on both sides.

impact27: > Using encoding='latin1' is required for unpickling NumPy arrays and instances of datetime, date and time pickled by Python 2.

I'll do that. There might be encoding problems but I don't really have a choice. (Anyway encoding problems is the reason why python2 still exists)

ccordoba12: Well, if there are encoding problems, `None` should be sent over the wire.

And thanks for bringing a better understanding about this!

impact27: Ok this is not a problem because I send the values in the 'buffer' part of the messages which is just data, not a python object such as `string` or `bytes`. BUT, the encoding issue is still present, so I will fix that.

impact27: `svalue = bytes(svalue, 'latin-1')` is only iseful to cast a string object to a bytes object, but the buffer will be opened correctly for any combination (as it is just `memory`).

impact27: > Ok, then this means we need to use the same encoding to serialize and deserialize on both sides.

We can't do that as we have no control of the encoding in python2. Basically the encoding will be: whatever the user decides, and the decoding will be 'latin-1', because otherwise numpy won't work. This means that anything that is encoded with utf-8 will be corrupted. (The previous behaviour was just crashing, so this is an improvment.) I don't think there is anything we can do. This kind of problem is exactly why python3 is a breaking update.

impact27: > Well, if there are encoding problems, `None` should be sent over the wire.

We won't know if there are encoding problems. The behaviour in python2 is : encode with whatever, decode with watever, hope it works. The behaviour in python3 is: encode with utf8, decode with utf8. If a python 3 string is sent to python 2, it will be shown as an unicode string (Which might be surprising, but at least will be decoded correctly).

spyder_kernels/comms/frontendcomm.py
impact27: Well it could decide to do something else instead. The problem is that this is an interface to call any function. If I call a function and expect that the function might throw an error, I need to add a `try` block, so I can handle the failure gracefully. This is the abstract interface used by all functions. For example, if I call `mydict[key]`, I want to get an exception if there is an error, so I can handle it. I wouldn't want to get a None back (because None might be the return value of a sucessful call.)

impact27: > What could the calling code do with a TimeoutError?
I mean if we don't want the users to report exceptions on github, why have a windows that ask them to do that? Are the errors happening between spyder-kernel and spyder less important than other errors?

impact27: The errors should be handled by the calling code as the calling code knows more about the situation than this code. The only things we knpw at this level are: "something went wrong somewhere". The normal way to handle that is to raise an error so the calling code can make a decision about it, not returning a None. Returning a none means that the function doesn't return anything, not that it failed.

spyder_kernels/comms/commbase.py
impact27: Well the user won't have to see these errors if they are handled correctly by the calling code. It the user sees this error, it means: something went wrong and the code did not handle it. This is true of any error, not just here. Otherwise why don't we log all the errors instead of having a pop-up windows showing them to the user?

spyder_kernels/comms/commbase.py
impact27: Yes, any code that calls a function that can throw an error needs to know what to do in cae of an error, or pass the error up in the stack. Here, 'up in the stack' means to the other side of the connection, so that is where this error goes.

spyder_kernels/comms/commbase.py
impact27: Here, I log the error instead of raising it because this is just a message, not a function call.

spyder_kernels/comms/frontendcomm.py
ccordoba12: > If I call a function and expect that the function might throw an error

Exactly, that's *if* you expect an error. If you don't, then the function should handle a None response (which is what I'm suggesting and what we already have in place).

> > What could the calling code do with a TimeoutError?

> I mean if we don't want the users to report exceptions on github, why have a windows that ask them to do that?

That's not the point. The point is that this kind of errors would heavily depend on the current user namespace, which would make them almost unreproducible for us (e.g. what if they start to appear after a week long session in which the user loaded a lot of data and did a lot of things with them?).

In that case, I don't want people to report errors that we won't be able to reproduce nor to fix. What's the point on allowing that?

> The only things we knpw at this level are: "something went wrong somewhere"

Nop, what we know at this point (and what you're reporting with your raise) is "the operation timed out". That's it.

And my proposal is "if things time out, then we should do nothing about it".

spyder_kernels/comms/commbase.py
ccordoba12: Ok, if errors are not ignored here (which I find easier), then they need to ignored on the frontend side, except for `get_value`.

spyder_kernels/comms/frontendcomm.py
impact27: >Exactly, that's if you expect an error. If you don't, then the function should handle a None response (which is what I'm suggesting and what we already have in place).

Well If the function should handle a `None` response then surely if expects something to not work? I know that this is how things work now but the code was specific to each function. Here as the code is general, and a function might want to know if an error happens, I need to raise an error. For example, if I want to have the other side do something, wait for that to be done, and then proceed, I can call a blocking function. If the blocking function doesn't return anything, I will get a None. Now I have a problem. Did the other side actually finished executing, or was there a timeout? I have no way to know.

> That's not the point. The point is that this kind of errors would heavily depend on the current user namespace, which would make them almost unreproducible for us (e.g. what if they start to appear after a week long session in which the user loaded a lot of data and did a lot of things with them?).

> In that case, I don't want people to report errors that we won't be able to reproduce nor to fix. What's the point on allowing that?

That is specific to the namespace related functions, and should be handled there. This code could be doing anything.

> And my proposal is "if things time out, then we should do nothing about it".

Maybe there is something better to do than nothing, but this function doesn't know that, so it should ask the calling function to figure something out. Maybe the best thing to do is return None? Maybe it is doing something else? The calling function knows about that, but this function doesn't, so the decision should be taken by the calling function.

This is a general interface and could be used by plugins, so there is no way of knowing what this interface will be used for. Assuming that returning None is always the best thing to do is a bit of a stretch. On the other hand, knowing that when you call a blocking function, it might timeout, and plan something seems more reasonable. Again, I am not sugesting that the user should see these errors, I just want the functions that call this interface to handle a timeout.

spyder_kernels/comms/commbase.py
impact27: Agreed

spyder_kernels/comms/commbase.py
ccordoba12: Ok, fair enough.

spyder_kernels/console/kernel.py
ccordoba12: I think pickled string are encoded as `latin-1` in Python 2. At least, that's what I found with a simple test in IPython:

```python
In [1]: import cloudpickle

In [2]: svalue = cloudpickle.dumps('ñ')

In [3]: svalue.decode()
---------------------------------------------------------------------------
UnicodeDecodeError                        Traceback (most recent call last)
<ipython-input-3-ab1c9e779743> in <module>()
----> 1 svalue.decode()

UnicodeDecodeError: 'ascii' codec can't decode byte 0x80 in position 0: ordinal not in range(128)

In [4]: svalue.decode('latin-1')
Out[4]: u'\x80\x02U\x02\xc3\xb1q\x00.'
```

spyder_kernels/comms/commbase.py
ccordoba12: > then they need to ignored on the frontend side

Sorry, scratch ~~frontend~~ above. It's really "the other side" (I forgot this is placed in CommBase).

> Agreed

Ok, I agree with this too. Please take care of correctly handling errors in the handlers present here and those in spyder-ide/spyder#9343.

spyder_kernels/comms/frontendcomm.py
ccordoba12: > Well If the function should handle a None response then surely if expects something to not work?

Yes, that's the idea. But your comment made me realize we're proposing two different solutions to the same problem: what to do if a blocking call fails? I propose returning None and you returning a TimeoutError.

That's sort of similar, but then this other comment

> This is a general interface and could be used by plugins, so there is no way of knowing what this interface will be used for.

made me realize that my proposed solution is inferior to yours because in the first case third-party developers would need to know that the response on failure is None, whereas in the second one they would naturally get an error.

So, I defer to your point here and I agree we must return an error and let callers handle it as they consider appropriate.

spyder_kernels/console/kernel.py
impact27: - Python 2 strings don't have encoding, they are basically `bytes` from python3. The encoding depends on where the strings come from. (File, keyboard, or other)
 - Python 3 string are UTF-8 encoded data.
 - If you want to display a string in Python2, you have to know what the encoding is. If you don't, you are out of luck, and will end up displaying computer gibbrish if you have non ascii letters. This is a major problem with python2, and the reason why python3 is a breaking update.
 - In Python3, you know the encoding, so no problem

- Now pickle want to serialise a list of strings. In Python2 pickle can not know anything about encoding, and therefore saves the string as is.
 - In Python3, pickle must create strings as unicode strings, but it has no clue what the encoding of the strings are (the encoding of each string could be different.) Therefore there is a few options:
    - Open all the strings as `bytes`. This is the correct way but then the user has to decode every single string that was pickled.
    - Open all the strings as `ascii`. Works for all ascii strings but fails as soon as a non ascii charater is found. This is the default for pickle.
    - Choose a random encoding and accept that the non-ascii letters might be gibbrish. This is what we do here.

ccordoba12:
Thanks a lot for your hard work on this @impact27!! I like this architecture a lot!

I was playing with it today and found a important shortcoming that I'd like to see addressed before merging: you're not handling response errors gracefully on either side of the comms. This is different from the long discussions we've had the last couple of days, so let me explain what I mean by this.

First, let's say I introduce a dummy error in `get_cwd` method of SpyderKernel, like this:

Selección_012

Then, let's run `cd` in the console to change the cwd:

Selección_013

After doing that, I got an error on Spyder's error report dialog with the following details:

Selección_014

As you can see, the error doesn't contain the actual one generated in the kernel. What appears there is the error in the `remote_set_cwd` method of ShellWidget:

Selección_015

when trying to emit a signal with the wrong data, which is the list of `[error, traceback]` that actually comes from the kernel:

> [ImportError('No module named foo',), [('/home/carlos/Projects/spyder/spyder-kernels/spyder_kernels/comms/commbase.py', 258, '_handle_remote_call', "buffer['call_kwargs'])"), ('/home/carlos/Projects/spyder/spyder-kernels/spyder_kernels/comms/commbase.py', 268, '_remote_callback', '*call_args, **call_kwargs)'), ('/home/carlos/Projects/spyder/spyder-kernels/spyder_kernels/console/kernel.py', 284, 'get_cwd', 'import foo')]]

----

So, instead of swallowing comm errors (which I was insisting we should do before), now I think we should properly log them through our error report dialog, so we can detect errors in our kernel handlers and fix them accordingly.

But in order to do that, we need to handle error responses in the comms when they arrive to the other side. So, if a kernel comm error is detected, then it should go directly to the frontend error dialog instead of calling the callback.

The only errors we can't log at the moment are the ones from kernel handlers that don't require a response (e.g. `set_value`). I really don't know where those errors are going at the moment (I introduced a dummy error in `set_value` of SpyderKernel but got nothing in Spyder).

ccordoba12:
(Sorry for closing this, my mistake).

spyder_kernels/console/kernel.py
ccordoba12: It seems you understand better the differences between Python 2 and 3 string representations than me! So I defer to your judgment here.

impact27:
It now sends the error from the kernel to the frontend where it is displayed in the error report dialog.

impact27:
By the way, you can create new callable kernel-side client from the frontend like this:
```
In [1] def get_cwd(): import os; os.getcwd()

In [2] get_ipython().kernel.frontend_comm.register_call_handler('set_cwd', get_cwd)
```

spyder_kernels/comms/commbase.py
```diff
@@ -0,0 +1,464 @@
...
+        """
+        if not self.is_open(comm_id):
+            raise CommError("The comm is not connected.")
+        import cloudpickle
+        msg_dict = {
+            'spyder_msg_type': spyder_msg_type,
+            'content': content,
+            'PY2': PY2,
+            }
+        buffers = [cloudpickle.dumps(data, protocol=PICKLE_PROTOCOL)]
```
ccordoba12: I think we're not handling correctly the case of a failure in serializing the requested data here. Instead of generating a traceback due to that (which is shown in our error dialog now), we need to catch `PicklingError` (or perhaps any exception?) and make `buffers` content return something that can be interpreted correctly on the other side.

This is quite similar to the way you're handling failures in `cloudpickle.loads` below.

spyder_kernels/comms/commbase.py
```diff
@@ -0,0 +1,464 @@
...
+        comm_id: int
+            the comm to send to. If None sends to all comms.
+        """
+        if not self.is_open(comm_id):
+            raise CommError("The comm is not connected.")
+        import cloudpickle
+        msg_dict = {
+            'spyder_msg_type': spyder_msg_type,
+            'content': content,
+            'PY2': PY2,
```
ccordoba12: Now that we have the chance to easily check if frontend and kernel are running different major Python versions, I think we can improve things by choosing Pickle's HIGHEST_PROTOCOL if both frontend and kernel are running in Python 3.

That would give us access to all serialization improvements in Python 3 (like the upcoming PEP 574)!

spyder_kernels/comms/commbase.py
```diff
@@ -0,0 +1,464 @@
...
+            The (JSONable) content of the message
+        data: any
+            Any object that is serializable by cloudpickle (should be most
+            things). Will arrive as cloudpickled bytes in `.buffers[0]`.
+        comm_id: int
+            the comm to send to. If None sends to all comms.
+        """
+        if not self.is_open(comm_id):
+            raise CommError("The comm is not connected.")
+        import cloudpickle
```
ccordoba12: This import is unnecessary here.

spyder_kernels/comms/commbase.py
impact27: I disagree: There are two kinds of errors:
 - Errors caused by the same side, which can be raised as is.
 - Errors caused by the other side, that have to be caught and send back.

Here, either the first point applies, or the exception is caught in `_handle_remote_call`.

spyder_kernels/comms/commbase.py
impact27: Good idea

spyder_kernels/comms/commbase.py
impact27: So for example if the failure happens because the frontend requested data that could not be pickled, the error generated here will be sent to the frontend. Is it not what you see?

impact27: By running this code, I just have one QWindows showing up.
```
try:
    1/0
except:
    import sys
    tb = sys.exc_info()[2]
var = [tb]
```

impact27: Screenshot 2019-08-09 at 16 47 13

ccordoba12: > Is it not what you see?

Yes, that's what I'm seeing. But I don't want users to report it in case they are trying to visualize a non-picklable object in the Variable Explorer.

A very simple example:

```python
In [1]: import sys

In [2]: f = sys._getframe()
```

If you double-click in `f` in the Variable Explorer, you'll get:

Selección_016

Showing the QMessageBox in the front is what we were doing before, and that's fine. But showing the error dialog is completely unnecessary because frames are not picklable.

impact27: I agree but I am not sure why you see the error dialog. It doesn't appear on my laptop.

impact27: Screenshot 2019-08-09 at 17 05 52

impact27: Also f is not in the variable explorer?

impact27: I would be templted to say that this is out of the scope of this PR.

impact27: What is the error and traceback in the  error dialog you have?

impact27: Or even simpler: you could open a new issue with the error dialog so I can see the traceback and other infos :p

ccordoba12: This was probably fixed with your switch to use Pickle HIGHEST_PROTOCOL, because I'm not seeing the error dialog shown anymore (only the QMessageBox).

So thanks for that!

spyder_kernels/comms/commbase.py
```diff
@@ -0,0 +1,527 @@
...
+            things). Will arrive as cloudpickled bytes in `.buffers[0]`.
+        comm_id: int
+            the comm to send to. If None sends to all comms.
+        """
+        if not self.is_open(comm_id):
+            raise CommError("The comm is not connected.")
+        msg_dict = {
+            'spyder_msg_type': spyder_msg_type,
+            'content': content,
+            'pickle protocol': self._pickle_protocol,
```
ccordoba12: Please remove the space in this key, i.e. rename it to `pickle_protocol`.

ccordoba12: Besides, please keep passing a key called `PY2` set to `True` if this is in Python 2 version. We could need that to make other adjustments in the future. Thanks!

ccordoba12:
Please also increase this value:

https://github.com/spyder-ide/spyder-kernels/blob/master/.codecov.yml#L12

to 5% to avoid your PR being flagged as failing by codecov.

PR #112, opened 2019-05-16
PR: Use the comms API to improve runcell and add a new debugcell command

### LINKED ISSUE #97: Can't repeat runcell from terminal

impact27:
This PR implements the changes needed by https://github.com/spyder-ide/spyder/pull/9014 to work. It is dependent on #111.

Fixes #97

Add a new builtin, `debugcell`, to debug a single cell

### DIFF
--- a/spyder_kernels/console/tests/test_console_kernel.py
+++ b/spyder_kernels/console/tests/test_console_kernel.py
@@ -354,45 +354,6 @@ if __name__ == '__main__':
         assert content['found']
 
 
-def test_runcell(tmpdir):
-    """Test the runcell command."""
-    # Command to start the kernel
-    cmd = "from spyder_kernels.console import start; start.main()"
-
-    with setup_kernel(cmd) as client:
-        # Write code with a cell to a file
-        code = u"result = 10; fname = __file__"
-        p = tmpdir.join("cell-test.py")
-        p.write(code)
-
-        # Attach cell_code to the IPython shell instance to simulate
-        # that the code was sent from Spyder's Editor
-        client.execute(u"get_ipython().cell_code = '{}'".format(code))
-        client.get_shell_msg(block=True, timeout=TIMEOUT)
-
-        # Execute runcell
-        client.execute(u"runcell('', r'{}')".format(to_text_string(p)))
-        client.get_shell_msg(block=True, timeout=TIMEOUT)
-
-        # Verify that the `result` variable is defined
-        client.inspect('result')
-        msg = client.get_shell_msg(block=True, timeout=TIMEOUT)
-        content = msg['content']
-        assert content['found']
-
-        # Verify that the `fname` variable is `cell-test.py`
-        client.inspect('fname')
-        msg = client.get_shell_msg(block=True, timeout=TIMEOUT)
-        content = msg['content']
-        assert "cell-test.py" in content['data']['text/plain']
-
-        # Verify that the `__file__` variable is undefined
-        client.inspect('__file__')
-        msg = client.get_shell_msg(block=True, timeout=TIMEOUT)
-        content = msg['content']
-        assert not content['found']
-
-
 def test_np_threshold(kernel):
     """Test that setting Numpy threshold doesn't make the Variable Explorer slow."""
 
--- a/spyder_kernels/console/tests/test_console_kernel.py
+++ b/spyder_kernels/console/tests/test_console_kernel.py
@@ -417,45 +417,6 @@ def test_runfile(tmpdir):
         assert content['found']
 
 
-def test_runcell(tmpdir):
-    """Test the runcell command."""
-    # Command to start the kernel
-    cmd = "from spyder_kernels.console import start; start.main()"
-
-    with setup_kernel(cmd) as client:
-        # Write code with a cell to a file
-        code = u"result = 10; fname = __file__"
-        p = tmpdir.join("cell-test.py")
-        p.write(code)
-
-        # Attach cell_code to the IPython shell instance to simulate
-        # that the code was sent from Spyder's Editor
-        client.execute(u"get_ipython().cell_code = '{}'".format(code))
-        client.get_shell_msg(block=True, timeout=TIMEOUT)
-
-        # Execute runcell
-        client.execute(u"runcell('', r'{}')".format(to_text_string(p)))
-        client.get_shell_msg(block=True, timeout=TIMEOUT)
-
-        # Verify that the `result` variable is defined
-        client.inspect('result')
-        msg = client.get_shell_msg(block=True, timeout=TIMEOUT)
-        content = msg['content']
-        assert content['found']
-
-        # Verify that the `fname` variable is `cell-test.py`
-        client.inspect('fname')
-        msg = client.get_shell_msg(block=True, timeout=TIMEOUT)
-        content = msg['content']
-        assert "cell-test.py" in content['data']['text/plain']
-
-        # Verify that the `__file__` variable is undefined
-        client.inspect('__file__')
-        msg = client.get_shell_msg(block=True, timeout=TIMEOUT)
-        content = msg['content']
-        assert not content['found']
-
-
 def test_np_threshold(kernel):
     """Test that setting Numpy threshold doesn't make the Variable Explorer slow."""
 
--- a/spyder_kernels/customize/spydercustomize.py
+++ b/spyder_kernels/customize/spydercustomize.py
@@ -847,7 +847,10 @@ def get_current_file_name():
     """Get the current file name."""
     try:
         return _frontend_request().current_filename()
-    except BaseException:
+    except Exception:
+        _print("This command failed to be executed because an error occurred"
+               " while trying to get the current file name from Spyder's"
+               " editor. The error was:\n\n")
         get_ipython().showtraceback(exception_only=True)
         return None
 
@@ -882,7 +885,7 @@ def runfile(filename=None, args=None, wdir=None, namespace=None,
     try:
         # Save the open files
         _frontend_request().save_files()
-    except BaseException:
+    except Exception:
         logger.debug("Could not save files before executing.")
 
     try:
@@ -993,7 +996,10 @@ def runcell(cellname, filename=None):
     try:
         # Get code from spyder
         cell_code = _frontend_request().run_cell(cellname, filename)
-    except BaseException:
+    except Exception:
+        _print("This command failed to be executed because an error occurred"
+               " while trying to get the cell code from Spyder's"
+               " editor. The error was:\n\n")
         get_ipython().showtraceback(exception_only=True)
         return
 
@@ -1060,7 +1066,7 @@ def cell_count(filename=None):
         # Get code from spyder
         cell_count = _frontend_request().cell_count(filename)
         return cell_count
-    except BaseException:
+    except Exception:
         etype, error, tb = sys.exc_info()
         raise etype(error)
 
--- a/spyder_kernels/customize/spydercustomize.py
+++ b/spyder_kernels/customize/spydercustomize.py
@@ -847,9 +847,8 @@ def get_current_file_name():
     """Get the current file name."""
     try:
         return _frontend_request().current_filename()
-    except Exception:
-        etype, error, _ = sys.exc_info()
-        get_ipython().showtraceback((etype, error, None))
+    except BaseException:
+        get_ipython().showtraceback(exception_only=True)
         return None
 
 
@@ -880,7 +879,7 @@ def runfile(filename=None, args=None, wdir=None, namespace=None,
     try:
         # Save the open files
         _frontend_request().save_files()
-    except Exception:
+    except BaseException:
         logger.debug("Could not save files before executing.")
 
     try:
@@ -979,9 +978,8 @@ def runcell(cellname, filename=None):
     try:
         # Get code from spyder
         cell_code = _frontend_request().run_cell(cellname, filename)
-    except Exception:
-        etype, error, tb = sys.exc_info()
-        get_ipython().showtraceback((etype, error, None))
+    except BaseException:
+        get_ipython().showtraceback(exception_only=True)
         return
 
     if not cell_code:
@@ -1047,7 +1045,7 @@ def cell_count(filename=None):
         # Get code from spyder
         cell_count = _frontend_request().cell_count(filename)
         return cell_count
-    except Exception:
+    except BaseException:
         etype, error, tb = sys.exc_info()
         raise etype(error)
 
--- a/spyder_kernels/customize/spydercustomize.py
+++ b/spyder_kernels/customize/spydercustomize.py
@@ -832,18 +832,45 @@ def _frontend_request(blocking=True):
         blocking=blocking, broadcast=False)
 
 
-def runfile(filename, args=None, wdir=None, namespace=None, post_mortem=False):
+def get_current_file_name():
+    """Get the current file name."""
+    try:
+        return _frontend_request().current_filename()
+    except Exception:
+        etype, error, _ = sys.exc_info()
+        get_ipython().showtraceback((etype, error, None))
+        return None
+
+
+def get_debugger(filename):
+    """Get a debugger for a given filename."""
+    debugger = pdb.Pdb()
+    filename = debugger.canonic(filename)
+    debugger._wait_for_mainpyfile = 1
+    debugger.mainpyfile = filename
+    debugger._user_requested_quit = 0
+    if os.name == 'nt':
+        filename = filename.replace('\\', '/')
+    return debugger, filename
+
+
+def runfile(filename=None, args=None, wdir=None, namespace=None,
+            post_mortem=False, is_pdb=False):
     """
     Run filename
     args: command line arguments (string)
     wdir: working directory
     post_mortem: boolean, whether to enter post-mortem mode on error
     """
+    if filename is None:
+        filename = get_current_file_name()
+        if filename is None:
+            return
     try:
         # Save the open files
         _frontend_request().save_files()
-    except (CommError, TimeoutError):
-        logger.debug("Could not send save files before executing.")
+    except Exception:
+        logger.debug("Could not save files before executing.")
 
     try:
         filename = filename.decode('utf-8')
@@ -894,18 +921,6 @@ def runfile(filename, args=None, wdir=None, namespace=None, post_mortem=False):
 builtins.runfile = runfile
 
 
-def get_debugger(filename):
-    """Get a debugger for a given filename."""
-    debugger = pdb.Pdb()
-    filename = debugger.canonic(filename)
-    debugger._wait_for_mainpyfile = 1
-    debugger.mainpyfile = filename
-    debugger._user_requested_quit = 0
-    if os.name == 'nt':
-        filename = filename.replace('\\', '/')
-    return debugger, filename
-
-
 def debugfile(filename, args=None, wdir=None, post_mortem=False):
     """
     Debug filename
@@ -913,6 +928,10 @@ def debugfile(filename, args=None, wdir=None, post_mortem=False):
     wdir: working directory
     post_mortem: boolean, included for compatiblity with runfile
     """
+    if filename is None:
+        filename = get_current_file_name()
+        if filename is None:
+            return
     debugger, filename = get_debugger(filename)
     debugger.run("runfile(%r, args=%r, wdir=%r)" % (filename, args, wdir))
 
@@ -935,6 +954,10 @@ def runcell(cellname, filename=None):
     filename : str
         Needed to allow for proper traceback links.
     """
+    if filename is None:
+        filename = get_current_file_name()
+        if filename is None:
+            return
     try:
         filename = filename.decode('utf-8')
     except (UnicodeError, TypeError, AttributeError):
@@ -945,11 +968,9 @@ def runcell(cellname, filename=None):
     try:
         # Get code from spyder
         cell_code = _frontend_request().run_cell(cellname, filename)
-        if not filename:
-            filename = _frontend_request().current_filename()
-    except (CommError, TimeoutError):
-        _print("--Run Cell Error--\n"
-               "Please use only with Spyder; ")
+    except Exception:
+        etype, error, tb = sys.exc_info()
+        get_ipython().showtraceback((etype, error, None))
         return
 
     if not cell_code:
@@ -979,6 +1000,23 @@ def runcell(cellname, filename=None):
 builtins.runcell = runcell
 
 
+def debugcell(cellname, filename=None):
+    """Debug a cell."""
+    if filename is None:
+        filename = get_current_file_name()
+        if filename is None:
+            return
+
+    debugger, filename = get_debugger(filename)
+    # The breakpoint might not be in the cell
+    debugger.continue_if_has_breakpoints = False
+    debugger.run("runcell({}, {}, is_pdb=True)".format(
+        repr(cellname), repr(filename)))
+
+
+builtins.debugcell = debugcell
+
+
 def cell_count(filename=None):
     """
     Get the number of cells in a file.
@@ -988,37 +1026,20 @@ def cell_count(filename=None):
     filename : str
         The file to get the cells from. If None, the currently opened file.
     """
+    if filename is None:
+        filename = get_current_file_name()
+        if filename is None:
+            raise RuntimeError('Could not get cell count from frontend.')
     try:
         # Get code from spyder
         cell_count = _frontend_request().cell_count(filename)
-        return int(cell_count)
-    except (CommError, TimeoutError):
-        _print("--Cell Count Error--\n"
-               "Please use only with Spyder; ")
-        return 0
+        return cell_count
+    except Exception:
+        etype, error, tb = sys.exc_info()
+        raise etype(error)
 
 
 builtins.cell_count = cell_count
-
-
-def debugcell(cellname, filename=None):
-    """Debug a cell."""
-    if not filename:
-        try:
-            filename = _frontend_request().current_filename()
-        except (CommError, TimeoutError):
-            pass
-
-    if not filename:
-        raise RuntimeError("Could not get the file name from Spyder.")
-
-    debugger, filename = get_debugger(filename)
-    # The breakpoint might not be in the cell
-    debugger.continue_if_has_breakpoints = False
-    debugger.run("runcell({}, {})".format(repr(cellname), repr(filename)))
-
-
-builtins.debugcell = debugcell
 
 
 #==============================================================================
--- a/spyder_kernels/customize/spydercustomize.py
+++ b/spyder_kernels/customize/spydercustomize.py
@@ -896,12 +896,6 @@ builtins.runfile = runfile
 
 def get_debugger(filename):
     """Get a debugger for a given filename."""
-    try:
-        # Save the open files so the debugging display is correct
-        _frontend_request().save_files()
-    except (CommError, TimeoutError):
-        logger.debug("Could not send save files before executing.")
-
     debugger = pdb.Pdb()
     filename = debugger.canonic(filename)
     debugger._wait_for_mainpyfile = 1
--- a/spyder_kernels/customize/spydercustomize.py
+++ b/spyder_kernels/customize/spydercustomize.py
@@ -313,6 +313,10 @@ class SpyderPdb(pdb.Pdb, object):  # Inherits `object` to call super() in PY2
 
     send_initial_notification = True
     starting = True
+    def __init__(self, *args, **kwargs):
+        """Init Pdb."""
+        self.continue_if_has_breakpoints = True
+        super(SpyderPdb, self).__init__(*args, **kwargs)
 
     def __init__(self, *args, **kwargs):
         """Init Pdb."""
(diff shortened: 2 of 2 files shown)


impact27:
This is ready for review once #111 is merged.

spyder_kernels/comms/kernelcomm.py
```diff
@@ -0,0 +1,105 @@
...
+# Licensed under the terms of the MIT License
+# (see spyder/__init__.py for details)
+
+"""
+In addition to the remote_call mechanism implemented in CommBase:
+ - Send a message to a debugging kernel
+"""
+import logging
+
+from qtpy.QtCore import QEventLoop, QObject, QTimer, Signal
```
ccordoba12: This is really problematic. This package should not depend on Qt because it would make it quite heavy and difficult to install.

Since this adds new functionality (i.e. sending a message to a debugging kernel), I'd say we should leave this for another PR for now, so we can discuss about it later.

spyder_kernels/comms/tests/test_comms.py
```diff
@@ -0,0 +1,158 @@
...
+    # Bypass the target system as this is not what is being tested
+    frontend_comm._comm_open(commB, {'content': None})
+
+    return (kernel_comm, frontend_comm)
+
+# =============================================================================
+# Tests
+# =============================================================================
+
+
```
ccordoba12: Please remove these blank lines.

spyder_kernels/console/tests/test_console_kernel.py
```diff
@@ -148,39 +170,27 @@ def test_get_var_properties(kernel):
...
-    assert "{'spyder_msg_type': 'data'}" in log_text
+    assert kernel.get_value(name) == 124
 
 
 def test_set_value(kernel):
     """Test setting the value of a variable."""
     name = 'a'
     execute = kernel.do_execute('a = 0', True)
-
+    PICKLE_PROTOCOL = 2
```
ccordoba12: PICKLE_PROTOCOL should be a constant that we can import from somewhere, as before.

spyder_kernels/console/tests/test_console_kernel.py
```diff
@@ -148,39 +170,27 @@ def test_get_var_properties(kernel):
...
     value = [cloudpickle.dumps(10, protocol=PICKLE_PROTOCOL)]
     PY2_frontend = False
     kernel.set_value(name, value, PY2_frontend)
     log_text = get_log_text(kernel)
     assert "'__builtin__': <module " in log_text
     assert "'__builtins__': <module " in log_text
-    assert "'_ih': ['']" in log_text
```
ccordoba12: Why did you need to remove `]` here?

spyder_kernels/console/tests/test_console_kernel.py
```diff
@@ -301,12 +311,6 @@ def test_output_from_c_libraries(kernel, capsys):
 libc.printf(('Hello from C\\n').encode('utf8'))
 """
 
-    # Without Wurlitzer there's not output generated
-    # by the kernel
-    reply = kernel.do_execute(code, True)
-    captured = capsys.readouterr()
-    assert captured.out == ''
```
ccordoba12: Why did you remove this? Are you loading `wurlitzer` unconditionally now?

spyder_kernels/console/tests/test_console_kernel.py
```diff
@@ -388,13 +405,32 @@ def test_runcell(tmpdir):
...
-        # Attach cell_code to the IPython shell instance to simulate
-        # that the code was sent from Spyder's Editor
-        client.execute(u"get_ipython().cell_code = '{}'".format(code))
-        client.get_shell_msg(block=True, timeout=TIMEOUT)
+        class Signal():
+            def connect(self, function):
+                self.function = function
+
+        # Fake Qt signal
+        iopub_recieved = Signal()
```
ccordoba12: recieved -> received

spyder_kernels/console/tests/test_console_kernel.py
```diff
@@ -388,13 +405,32 @@ def test_runcell(tmpdir):
...
+        client.iopub_channel.message_received = iopub_recieved
+        # Open comm
+        comm_manager = CommManager(client)
+        kernel_comm = KernelComm()
+        kernel_comm._register_comm(comm_manager.new_comm('spyder_api'))
+
+        def process_msg():
+            msg = {'msg_type': None}
+            while msg['msg_type'] != 'comm_msg':
+                msg = client.get_iopub_msg(block=True, timeout=TIMEOUT)
```
ccordoba12: Why does this need to be a blocking call now?

spyder_kernels/console/tests/test_console_kernel.py
```diff
@@ -508,6 +544,8 @@ def test_turtle_launch(tmpdir):
 
         # Run code
         client.execute("runfile(r'{}')".format(to_text_string(p)))
+        time.sleep(1)
+        client.input('')
```
ccordoba12: Why do you need to call `input` here?

spyder_kernels/console/tests/test_console_kernel.py
```diff
@@ -524,6 +562,8 @@ def test_turtle_launch(tmpdir):
 
         # Run code again
         client.execute("runfile(r'{}')".format(to_text_string(p)))
+        time.sleep(1)
+        client.input('')
```
ccordoba12: Also, why the call to `input` here?

spyder_kernels/customize/spydercustomize.py
```diff
@@ -347,10 +347,17 @@ def set_spyder_breakpoints(self, breakpoints):
             # Do 'continue' if the first breakpoint is *not* placed
             # where the debugger is going to land.
             # Fixes issue 4681
-            if breaks and lineno != breaks[0] and osp.isfile(fname):
-                get_ipython().kernel.pdb_continue()
+            if self.continue_if_has_breakpoints and \
+                    breaks and \
+                    lineno < breaks[0] and \
+                    osp.isfile(fname):
```
ccordoba12: Please use parenthesis instead of line continuations here, like this:

```python
if (self.continue_if_has_breakpoints and
        breaks and
        lineno < breaks[0] and
        osp.isfile(fname)):
    ...
```

spyder_kernels/customize/spydercustomize.py
```diff
@@ -347,10 +347,17 @@ def set_spyder_breakpoints(self, breakpoints):
...
+                try:
+                    get_ipython().kernel.pdb_continue()
+                except (CommError, TimeoutError):
+                    logger.debug("Could not send continue to frontend.")
```
ccordoba12: Please change this to

    Could not send a Pdb continue call to the frontend.

spyder_kernels/customize/spydercustomize.py
```diff
@@ -303,16 +285,34 @@ def _patched_preparation_data(name):
...
         """Ask Spyder for breakpoints before the first prompt is created."""
-        if self.starting:
-            get_ipython().kernel._ask_spyder_for_breakpoints()
+        try:
+            _frontend_request(blocking=False).set_debug(True)
+            if self.starting:
+                breakpoints = _frontend_request().get_breakpoints()
+                self.set_spyder_breakpoints(breakpoints)
+        except (CommError, TimeoutError):
+            logger.debug("Could not send preloop to frontend.")
```
ccordoba12: I think this should be

    Could not get breakpoint from the frontend

because that's what's really the problem here.

spyder_kernels/comms/kernelcomm.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def __init__(self):
+        super(KernelComm, self).__init__()
+
+        self.kernel_client = None
+        self._debugging = False
+        self._wait_list = {}
+
+        self.register_call_handler(
+            'set_debug', self._handle_debug_state)
```
ccordoba12: This should be `'set_debug_state'` because the meaning of `'set_debug'` is not so clear.

spyder_kernels/customize/spydercustomize.py
```diff
@@ -303,16 +285,34 @@ def _patched_preparation_data(name):
...
+
+
+    def postloop(self):
+        """Notifies spyder that the loop has ended."""
+        try:
+            _frontend_request(blocking=False).set_debug(False)
+        except (CommError, TimeoutError):
+            logger.debug("Could not send postloop to frontend.")
```
ccordoba12: This is confusing because `postloop` is set in the kernel (and you're doing it below this line).

Are you really informing the frontend that the debug session finished instead? If that's the case, why is that now necessary?

spyder_kernels/customize/spydercustomize.py
```diff
@@ -347,10 +347,17 @@ def set_spyder_breakpoints(self, breakpoints):
...
+                    breaks and \
+                    lineno < breaks[0] and \
+                    osp.isfile(fname):
+                try:
+                    get_ipython().kernel.pdb_continue()
+                except (CommError, TimeoutError):
+                    logger.debug("Could not send continue to frontend.")
 
     def notify_spyder(self, frame):
+        """Send state to spyder."""
```
ccordoba12: Send *kernel* state to the frontend.

spyder_kernels/customize/spydercustomize.py
```diff
@@ -374,7 +381,10 @@ def notify_spyder(self, frame):
         # Publish Pdb state so we can update the Variable Explorer
         # and the Editor on the Spyder side
         kernel._pdb_step = step
-        kernel.publish_pdb_state()
+        try:
+            kernel.publish_pdb_state()
+        except (CommError, TimeoutError):
+            logger.debug("Could not send pdb state to frontend.")
```
ccordoba12: Could not send *Pdb* state to *the* frontend.

spyder_kernels/customize/spydercustomize.py
```diff
@@ -473,7 +483,8 @@ def reset(self):
 #     specific behaviour desired?)
 @monkeypatch_method(pdb.Pdb, 'Pdb')
 def postcmd(self, stop, line):
-    if "_set_spyder_breakpoints" not in line:
+    if "_set_spyder_breakpoints" not in line and \
+            '!get_ipython().kernel' not in line:
```
ccordoba12: Please use parenthesis instead of a continuation line here.

spyder_kernels/customize/spydercustomize.py
```diff
@@ -823,7 +846,7 @@ def runfile(filename, args=None, wdir=None, namespace=None, post_mortem=False):
             ipython_shell = get_ipython()
             ipython_shell.run_cell_magic('cython', '', f.read())
     else:
-        execfile(filename, namespace)
+        get_ipython().safe_execfile(filename, namespace)
```
ccordoba12: Wow, this is a big change! Please revert it because it has nothing to do with this PR.

spyder_kernels/customize/spydercustomize.py
```diff
@@ -839,7 +862,40 @@ def runfile(filename, args=None, wdir=None, namespace=None, post_mortem=False):
...
+
+
+def debugfile(filename, args=None, wdir=None, post_mortem=False):
+    """
+    Debug filename
+    args: command line arguments (string)
+    wdir: working directory
+    post_mortem: boolean, included for compatiblity with runfile
+    """
+
```
ccordoba12: Please remove this blank line.

spyder_kernels/customize/spydercustomize.py
```diff
@@ -862,46 +917,92 @@ def runcell(cellname, filename):
...
 
     # Trigger `post_execute` to exit the additional pre-execution.
     # See Spyder PR #7310.
     ipython_shell.events.trigger('post_execute')
 
-    ipython_shell.run_cell(cell_code)
-    namespace.pop('__file__')
-    del ipython_shell.cell_code
+    try:
+        exec(compile(cell_code, filename, 'exec'), namespace)
```
ccordoba12: I don't like this. I think we should run `ipython_shell.run_cell`, as before.

spyder_kernels/customize/spydercustomize.py
```diff
@@ -862,46 +917,92 @@ def runcell(cellname, filename):
...
+        namespace['__file__'] = namespace_filename
+    else:
+        namespace.pop('__file__')
 
 
 builtins.runcell = runcell
 
 
-def debugfile(filename, args=None, wdir=None, post_mortem=False):
+def cellcount(filename=None):
```
ccordoba12: Why is this function needed? It doesn't seem that you're using it in this PR.

impact27:
> My main concern right now is that spyder-kernels shouldn't depend on Qt because it'd make it very heavy and sort of difficult to install. So I think you should remove the functionality that depends on Qt for now, so we can leave it for another PR (it could be optional instead of mandatory, so if Qt is not available, then we could simply skip it).

The dependency on Qt is specific to `spyder_kernels/comms/kernelcomm.py`. This file might need to be displaced to `spyder` instead of `spyder_kernels`. Indeed, it contains the frontend side of the communication. I put it there because:
 - I need this for tests
 - It kind of makes sense to keep all the comms part together?

The dependency on Qt of this file can not really be removed, but then again it is not really a file that is part of spyder_kernels. What do you think should be done?

spyder_kernels/comms/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+# (see spyder_kernels/__init__.py for details)
+# -----------------------------------------------------------------------------
+
+"""
+API to communicate between the spyder IDE and the spyder kernel.
+It uses IPython Comms for messaging. The messages are sent by calling an
+arbitrary function, with the limitation that the arguments have to be
+picklable. If the function must return, the call must be blocking.
+
+In addition, the frontend can interrupt the kernel to process the message sent.
```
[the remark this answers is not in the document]
impact27: Well this is a bit hard. Basically the kernel will get messages, but it is not checking them while the execution takes place. Therefore the kernel needs to be interrupted to process the messages. This is because coroutines are used instead of threads in ipykernel.
While debugging, this is done by sending a command to pdb:
`u"!get_ipython().kernel.frontend_comm.remote_call(blocking=True).pong()"
For general execution, the solution I found was to use signals (https://github.com/spyder-ide/spyder-kernels/pull/117), but this only work in unix. Another solution must be found for windows (or any platform that doesn't have `SIGUSR1`).

ccordoba12:
> The dependency on Qt is specific to spyder_kernels/comms/kernelcomm.py. This file might need to be displaced to spyder instead of spyder_kernels. Indeed, it contains the frontend side of the communication

Ok, thanks for the clarification. So those are the comms opened by the frontend on the kernel?

> What do you think should be done?

I think we should move `kernelcomm.py` to Spyder and also move the tests that depend on it there (you can move them to PR spyder-ide/spyder#9343). It's sort of the same situation that we have with the Variable Explorer settings: they live here but we need to use and set them in the Spyder side.

spyder_kernels/comms/commbase.py
```diff
@@ -0,0 +1,372 @@
...
+        self._comm = None
+        self._message_handlers = {}
+        self._remote_call_handlers = {}
+        self._call_reply_dict = {}
+        self._register_message_handler(
+            'remote_call', self._handle_remote_call)
+        self._register_message_handler(
+            'blocking_call_reply', self._handle_blocking_call_reply)
+        self.register_call_handler('ping', lambda: self.remote_call().pong())
+        self.register_call_handler('pong', lambda: None)
```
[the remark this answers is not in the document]
impact27: Well not only. If I want to send a non-comms message to the kernel while debugging (to get autocomplete for example), I need a way to interrupt the kernel. The easiest way is to make an interrupting call to the kernel. Another example is if I want to wait until the kernel replies, I want to make a blocking call. For these examples, `ping` and `pong` are handy "do nothing" functions to be used.

spyder_kernels/comms/commbase.py
```diff
@@ -0,0 +1,372 @@
...
+        self._remote_call_handlers[call_name] = handler
+
+    def remote_call(self, **settings):
+        """Get a handler for remote calls."""
+        return RemoteCallFactory(self, **settings)
+
+    # ---- Private -----
+    def _send_message(self, spyder_msg_type, content=None, data=None):
+        """
+        Publish custom messages to the Spyder kernel.
```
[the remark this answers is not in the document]
impact27: Well commbase is the base for both sides of the communication.

spyder_kernels/comms/commbase.py
```diff
@@ -0,0 +1,372 @@
...
+        """
+        settings = call_dict['settings']
+
+        if 'blocking' not in settings or not settings['blocking']:
+            return
+
+        if 'timeout' in settings:
+            timeout = settings['timeout']
+        else:
+            timeout = 3
```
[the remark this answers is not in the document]
goanpeca: Maybe add the unit as a comment?

```
timeout = 3  # Seconds
```

spyder_kernels/comms/commbase.py
```diff
@@ -0,0 +1,372 @@
...
+                msg_dict, buffer, load_exception)
+            return
+
+        logger.debug("No such spyder message type: %s" % spyder_msg_type)
+
+    def _handle_remote_call(self, msg, buffer, load_exception):
+        """Handle a remote call."""
+        msg_dict = msg['content']
+        if load_exception:
+            raise load_exception
```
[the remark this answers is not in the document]
impact27: If something doesn't happen, it can be frustrating for the user. As a user, I would rather know that something went wrong and be able to report it rather than not understanding why something is not happening.

spyder_kernels/comms/commbase.py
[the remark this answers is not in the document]
impact27: yes

spyder_kernels/comms/kernelcomm.py
impact27: This is the frontend -> kernel communication file. I will move that to spyder. Where should I put it?

spyder_kernels/console/tests/test_console_kernel.py
impact27: I can't remember exactly, I will investigate.

spyder_kernels/console/tests/test_console_kernel.py
impact27: I had problems with wurlitzer and I was loading it to avoid exceptions, but I fixed the problems in wurlitzer, so I am not sure if this part is still a problem.

spyder_kernels/console/tests/test_console_kernel.py
impact27: I am not sure I understand how this could not be blocking, I need to finish the test before the end of the function?

spyder_kernels/customize/spydercustomize.py
impact27: I want to know if the kernel is debugging or not, so I know if I need to use pdb to fetch the new messages. I use `u"!get_ipython().kernel.frontend_comm.remote_call(blocking=True).pong()"`, which only works while debugging.

impact27: Aldo this could be useful in spyder to make decisions on debugging. Now in the spyder code, `_reading` is taken as meaning `_reading_pdb` in many places, but you could have a call to `input()` where it might not be true. It wold therefore be useful to have a way of telling if debugging is really taking place.

spyder_kernels/customize/spydercustomize.py
impact27: I separated that to https://github.com/spyder-ide/spyder-kernels/pull/134.

Could you have a look at this PR? If you thing these changes are good I won't need to change that here.

spyder_kernels/customize/spydercustomize.py
impact27: Also in #134.

spyder_kernels/customize/spydercustomize.py
impact27: The new runcell function accept either a name or an index. I think this is better for axample in the following example:
```
print('a')
# %% Cell 1
print('b')
```
The first cell doesn't have a name, which is problematic. If I can call it by index as a fallback, I can just call `runcell(0)`. 

Now that I can call cells by index, I can write a script to run all the cells (because why not). But to do that, I need a way to know the number of cells. this is what `cellcount` does.

spyder_kernels/comms/__init__.py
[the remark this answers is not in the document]
ccordoba12: Ok, thanks for the explanation. It's not really critical to have that functionality yet.

spyder_kernels/comms/commbase.py
```diff
@@ -0,0 +1,372 @@
...
+PICKLE_PROTOCOL = 2
+
+class CommError(RuntimeError):
+    pass
+
+
+class CommBase(object):
+    """
+    Class with the necessary attributes and methods to handle
+    communications with a console.
```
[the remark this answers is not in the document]
impact27: Yes but this is a wrapper around jupyter comms, not the comms themeselves.

spyder_kernels/comms/tests/test_comms.py
impact27: Should I not leave two white spaces between two functions?

impact27:
All these functions:
```
            'set_breakpoints': self.set_spyder_breakpoints,
            'get_value': self.get_value,
            'load_data': self.load_data,
            'save_namespace': self.save_namespace,
            'is_defined': self.is_defined,
            'get_doc': self.get_doc,
            'get_source': self.get_source,
            'set_value': self.set_value,
            'remove_value': self.remove_value,
            'copy_value': self.copy_value,
            'set_cwd': self.set_cwd,
            'get_cwd': self.get_cwd,
            'update_cwd': self.update_cwd,
            'get_syspath': self.get_syspath,
            'get_env': self.get_env,
            'close_all_mpl_figures': self.close_all_mpl_figures,
            'show_mpl_backend_errors': self.show_mpl_backend_errors,
            'update_namespace_view': self.update_namespace_view,
            'set_namespace_view_settings': self.set_namespace_view_settings,
            'update_var_properties': self.update_var_properties,
```
Are now transmitted through comms. If I am to transfer the tests that depend on `kernelcomm` to spyder, that would mean moving almost all tests to spyder. Can I have spyder as a test dependency?

ccordoba12:
> Can I have spyder as a test dependency?

No, that would be a mess. The real tests for the interaction between Frontend and Kernel Comms are the ones already present in Spyder (e.g. getting a value from the kernel, setting a value in the kernel, runcell, etc, etc).

Therefore, please remove the tests you added here that require `KernelComm` (with the excepction of test_runcell, I think we could move them to Spyder).

ccordoba12:
**Note**: Please don't forget to implement my suggested changes here in PR #111.

ccordoba12:
Also, am I missing something or after this PR the kernel won't be able to work without Comms?

spyder_kernels/console/tests/test_console_kernel.py
impact27: from this line (not something I added): https://github.com/spyder-ide/spyder-kernels/blob/5496e4596dabda05d8583e2533fcdb14ebde2c9c/spyder_kernels/console/start.py#L126
I would expect wurlitzer to be loaded inconditionnaly?

ccordoba12:
I mean, without the frontend opening comms in the kernel and vice-versa?

impact27:
Yes it would still work.

spyder_kernels/console/tests/test_console_kernel.py
impact27: I added it back

spyder_kernels/console/tests/test_console_kernel.py
impact27: removed

spyder_kernels/console/tests/test_console_kernel.py
impact27: removed

spyder_kernels/customize/spydercustomize.py
ccordoba12: I'll take a look at this again once #111 is merged.

ccordoba12:
> Yes it would still work.

Correct. I tried this with external kernels and it's working was expected.

ccordoba12:
@impact27, let's continue with this one and it's corresponding PR in Spyder. The first step is for you to remove or squash all merge commits you have in this branch and leave a single merge with master. Or you can also rebase your initial commits (i.e. without the merges) on top of master.

impact27:
I would suggest https://github.com/spyder-ide/spyder/pull/9966 instead, then this one, as it is simpler.

spyder_kernels/console/tests/test_console_kernel.py
```diff
@@ -354,45 +354,6 @@ def f(x):
         assert content['found']
 
 
-def test_runcell(tmpdir):
```
ccordoba12: Is it possible to add a new test that replaces this one?

ccordoba12:
Ok, no problem.

spyder_kernels/console/tests/test_console_kernel.py
impact27: Well this test was not removed, it was just transferred to spyder. This makes sense because runcell doesn't work without a spyder frontend.

spyder_kernels/customize/spydercustomize.py
```diff
@@ -313,6 +313,10 @@ class SpyderPdb(pdb.Pdb, object):  # Inherits `object` to call super() in PY2
 
     send_initial_notification = True
     starting = True
+    def __init__(self, *args, **kwargs):
```
ccordoba12: Please add a blank line above this one.

spyder_kernels/customize/spydercustomize.py
```diff
@@ -917,46 +953,88 @@ def runcell(cellname, filename):
...
-    except AttributeError:
-        _print("--Run Cell Error--\n"
-               "Please use only through Spyder's Editor; "
-               "shouldn't be called manually from the console")
+        # Get code from spyder
+        cell_code = _frontend_request().run_cell(cellname, filename)
+        if not filename:
+            filename = _frontend_request().current_filename()
+    except (CommError, TimeoutError):
+        if hasattr(ipython_shell, "cell_code"):
```
ccordoba12: I think this should be removed. We should only use a single method in `runcell`.

spyder_kernels/customize/spydercustomize.py
```diff
@@ -917,46 +953,88 @@ def runcell(cellname, filename):
...
+        namespace['__file__'] = namespace_filename
+    else:
+        namespace.pop('__file__')
 
 
 builtins.runcell = runcell
 
 
-def debugfile(filename, args=None, wdir=None, post_mortem=False):
+def cellcount(filename=None):
```
ccordoba12: Let's rename this to `cell_count`.

spyder_kernels/customize/spydercustomize.py
```diff
@@ -917,46 +953,88 @@ def runcell(cellname, filename):
...
     """
-    Debug filename
-    args: command line arguments (string)
-    wdir: working directory
-    post_mortem: boolean, included for compatiblity with runfile
+    Get the number of cells for a file.
```
ccordoba12: for -> in

spyder_kernels/customize/spydercustomize.py
```diff
@@ -917,46 +953,88 @@ def runcell(cellname, filename):
...
+
+    Parameters
+    ----------
+    filename : str
+        Needed to allow for proper traceback links.
```
ccordoba12: This docstring is wrong. Please fix it.

spyder_kernels/customize/spydercustomize.py
ccordoba12: Also, what happens when you pass None as `filename` here?

spyder_kernels/customize/spydercustomize.py
```diff
@@ -917,46 +953,88 @@ def runcell(cellname, filename):
...
+def debugcell(cellname, filename=None):
+    """Debug a cell."""
+    if not filename:
+        try:
+            filename = _frontend_request().current_filename()
+        except (CommError, TimeoutError):
+            pass
+
+    if not filename:
+        raise RuntimeError("Could not get the filename from Spyder.")
```
ccordoba12: filename -> file name

spyder_kernels/customize/spydercustomize.py
ccordoba12: This is still unaddressed.

spyder_kernels/customize/spydercustomize.py
impact27: From shell.get_editor: 
```
if not filename:
            return editorstack.get_current_editor()
```

ccordoba12:
Why did you merged your `better_tracebacks` branch here? I thought that was part of another PR and that this PR was ready.

ccordoba12:
I'd prefer that you revert that change and leave it for later.

impact27:
Pdb doesn't stop unless we are in the correct file, because:
```
    debugger._wait_for_mainpyfile = 1
    debugger.mainpyfile = filename
```
Without `better_tracebacks`, the file name is not attached to the cell, and therefore debugcell never breaks and therefore never sets the breakpoints.

ccordoba12:
Ok, now this is a mess because you're mixing the improvements to `runcell` with the addition of `debugcell`.

So please remove the addition of `debugcell` from this PR and add it to PR #134 (and forget the last comment I post in PR #134). For that I think you need to remove the last five commits here. Then please remove `debugcell`, but you leave the improvements to `debugfile`.

Sorry to make you do more things, but PRs should be focused in solving one problem at a time. That makes my life as a reviewer way easier (as I said before, I thought this PR was ready to be merged).

ccordoba12:
Ok, this is also very confusing now. Please remove the last five commits from this PR (probably you need to reapply the one called "Fix cell_count type") and move the addition of `debugcell` on this side to another PR.

ccordoba12:
Thanks @impact27 for the extra effort!! This now requires a merge with master to fix the merge conflicts that appeared.

spyder_kernels/customize/spydercustomize.py
```diff
@@ -920,59 +974,98 @@ def runcell(cellname, filename):
...
-    debugger.run("runfile(%r, args=%r, wdir=%r)" % (filename, args, wdir))
+    if filename is None:
+        filename = get_current_file_name()
+        if filename is None:
+            raise RuntimeError('Could not get cell count from frontend.')
+    try:
+        # Get code from spyder
+        cell_count = _frontend_request().cell_count(filename)
+        return cell_count
+    except BaseException:
```
ccordoba12: Why `BaseException`?

spyder_kernels/customize/spydercustomize.py
```diff
@@ -920,59 +974,98 @@ def runcell(cellname, filename):
...
-               "shouldn't be called manually from the console")
+        # Get code from spyder
+        cell_code = _frontend_request().run_cell(cellname, filename)
+    except BaseException:
+        get_ipython().showtraceback(exception_only=True)
         return
 
+    if not cell_code:
+        # Nothing to execute
+        return
```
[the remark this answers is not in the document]
impact27: yes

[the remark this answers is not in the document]
ccordoba12: Great, thanks!

PR #113, opened 2019-05-17
PR: fix traceback in python3

impact27:
Running a file where an error happens, I get a few unwanted lines in the traceback.
For example, with a file that says: 
```
raise RuntimeError
```
### Before:
Screenshot 2019-05-17 at 15 13 12
### After:
Screenshot 2019-05-17 at 15 13 33


impact27:
This only works in python3, but I guess it should not be too difficult to adapt for python2.

impact27:
closing because I need this code in the other PR

PR #117, opened 2019-06-13
[Depends on #166] PR: Use unix signal to interrupt kernel

impact27:
This PR is built on top of PR #111  (and PR #112 but that can be removed, it is just easier for me.)
The problem I am trying to solve is the following:
 - If the kernel is executing something, it will not process incoming messages, because the ioloop is busy executing the code.
 - While debugging for example, this mean I can not set new breakpoints except if the debugging stopped (So no "Stop here, wherever 'here' is" possibility, or "oups, I forgot to put a breakopint here.")

ipykernel is using coroutines, which are not very compatible with threads. Therefore, another method must be used to do things "In parallel". The idea is to use unix signals (so no support for windows unfortunately). I don't want to interrupt anywhere in the `ioloop` code as it might cause problem, so the signal will abort unless:
 - `do_execute` is running
 - There is only one current call to `dispatch_shell` (just in case `do_execute` is calling `kernel.do_one_iteration()`)

An alternative would be to run the blocking code in another thread, but that would require modifying the ipykernel mechanisms.

### DIFF
--- a/spyder_kernels/comms/commbase.py
+++ b/spyder_kernels/comms/commbase.py
@@ -146,7 +146,6 @@ class CommBase(object):
         # Lists of reply numbers
         self._reply_inbox = {}
         self._reply_waitlist = {}
-        self._pickle_protocol = DEFAULT_PICKLE_PROTOCOL
 
         self._register_message_handler(
             'remote_call', self._handle_remote_call)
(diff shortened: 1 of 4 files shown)


impact27:
test is hanging

impact27:
Superseded by #169

PR #119, opened 2019-06-30
setup: add two missing requirements

sebix:
matplotlib is required in the tests since the last version 0.5.0 and mock was also missing

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,9 +39,9 @@ REQUIREMENTS = ['ipykernel>=4.8.2',
                 'pyzmq>=17',
                 'jupyter-client>=5.2.3',
                 'cloudpickle',
-                'wurlitzer;platform_system!="Windows"',
-                'mock',
-                'matplotlib']
+                'wurlitzer;platform_system!="Windows"']
+with io.open('requirements/tests.txt', encoding='utf-8') as f:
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -39,7 +39,9 @@ def get_version(module='spyder_kernels'):
                 'pyzmq>=17',
                 'jupyter-client>=5.2.3',
                 'cloudpickle',
-                'wurlitzer;platform_system!="Windows"']
+                'wurlitzer;platform_system!="Windows"',
+                'mock',
+                'matplotlib']
```
ccordoba12: These are not package requirements, but test ones. So what you need to do is add an `extras_require` kwarg to `setup` below and list all our test requirements there (not just these two).

Our test requirements are listed here:

https://github.com/spyder-ide/spyder-kernels/blob/master/requirements/tests.txt

ccordoba12:
I'm closing this because what's done against master doesn't affect our 0.x branch (where 0.5.0 was tagged).

I'll ask one of our contributors to fix this for you. Thanks for understanding.

sebix:
It would have been possible to change the target branch...

ccordoba12:
Sure, but you wouldn't be able to use `requirements/tests.txt` because it's not present in 0.x

PR #122, opened 2019-07-02
PR: Add tests requirements to setup.py

### LINKED ISSUE #121: Add test requirements to setup.py
Please follow [the way](https://github.com/spyder-ide/spyder/blob/master/setup.py#L241-L255) Spyder defines its test requirements through `extras_require` to declare our tests dependencies in the `setup.py` file placed here.

Since in `spyder_kernels` we are using setuptools directly, simply declare a constant called `TEST_REQUIREMENTS`, add the list of test packages to it and finally pass that to `extras_require` like this:

```python
setup(
    name='spyder-kernels',
    version=get_version(),
    ...
    install_requires=REQUIREMENTS,
    extras_require={'test': TEST_REQUIREMENTS}
    ...
)
```

Thanks!

steff456:
Fix #121

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -64,7 +64,7 @@ setup(
     long_description_content_type='text/markdown',
     packages=find_packages(exclude=['docs']),
     install_requires=REQUIREMENTS,
-    extras_requires={'test': TEST_REQUIREMENTS},
+    extras_require={'test': TEST_REQUIREMENTS},
     include_package_data=True,
     classifiers=['Development Status :: 5 - Production/Stable',
                  'Framework :: Jupyter',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -56,6 +64,7 @@ def get_version(module='spyder_kernels'):
     long_description_content_type='text/markdown',
     packages=find_packages(exclude=['docs']),
     install_requires=REQUIREMENTS,
+    extras_requires={'test': TEST_REQUIREMENTS},
```
ccordoba12: This needs to be `extras_require`, not `extras_requires`.

PR #124, opened 2019-07-05
PR: Change Variable Explorer settings from exclude_* to include_*

dalthviz:
Part of spyder-ide/spyder#9697


ccordoba12:
Closing this one because we're thinking in a different approach to the problem @dalthviz was trying to solve here.

PR #128, opened 2019-07-21
PR: Fix deprecated import

oscargus:
Just a minor fix, but required for Python 3.8 support.


oscargus:
Seems like the removal was postponed until 3.9 (as can be seen when running 3.8).

Issue #91, opened 2019-03-03
KeyError when running "%reset -f" programmatically


ccordoba12:
An error is generated when running `%reset -f` programmatically, i.e. like this

```python
from IPython import get_ipython
get_ipython().magic('reset -f')
```

The error is the following:

```python-traceback
Traceback (most recent call last):

  File "<ipython-input-1-537800c38694>", line 1, in <module>
    runfile('/home/carlos/.config/spyder-py3/temp.py')

  File "/home/carlos/Projects/spyder/spyder-kernels/spyder_kernels/customize/spydercustomize.py", line 807, in runfile
    namespace.pop('__file__')

KeyError: '__file__'
```

PR #134, opened 2019-08-03
PR: Improve the display of tracebacks and better handle namespace and __file__ during execution

impact27:
This is the companion of PR spyder-ide/spyder#10047.

----

# Before:

## runfile:

before

## runcell:

before_runcell

# After:

## runfile:

after

Less internal levels

## runcell:

after_runcell
The file name is correct.

### DIFF
--- a/spyder_kernels/customize/namespace_manager.py
+++ b/spyder_kernels/customize/namespace_manager.py
@@ -63,8 +63,6 @@ class NamespaceManager(object):
         if self._previous_filename:
             self.namespace['__file__'] = self._previous_filename
         elif '__file__' in self.namespace:
-            # Avoid error when running `%reset -f` programmatically
-            # See issue spyder-ide/spyder-kernels#91
             self.namespace.pop('__file__')
         if self._previous_main:
             sys.modules['__main__'] = self._previous_main
--- a/spyder_kernels/customize/namespace_manager.py
+++ b/spyder_kernels/customize/namespace_manager.py
@@ -1,0 +1,72 @@
+#
+# Copyright (c) 2009- Spyder Kernels Contributors
+#
+# Licensed under the terms of the MIT License
+# (see spyder_kernels/__init__.py for details)
+
+import sys
+
+from IPython.core.getipython import get_ipython
+
+
+def _get_globals():
+    """Return current namespace"""
+    ipython_shell = get_ipython()
+    return ipython_shell.user_ns
+
+
+class NamespaceManager(object):
+    """
+    Get a namespace and set __file__ to filename for this namespace.
+
+    The namespace is either namespace, the current namespace if
+    current_namespace is True, or a new namespace.
+    """
+
+    def __init__(self, filename, namespace=None, current_namespace=False):
+        self.filename = filename
+        self.namespace = namespace
+        self.current_namespace = current_namespace
+        self._previous_filename = None
+        self._previous_main = None
+        self._reset_main = False
+
+    def __enter__(self):
+        """
+        Prepare the namespace.
+        """
+        # Save previous __file__
+        if self.namespace is None:
+            if self.current_namespace:
+                self.namespace = _get_globals()
+            else:
+                ipython_shell = get_ipython()
+                main_mod = ipython_shell.new_main_mod(
+                    self.filename, '__main__')
+                self.namespace = main_mod.__dict__
+                # Needed to allow pickle to reference main
+                if '__main__' in sys.modules:
+                    self._previous_main = sys.modules['__main__']
+                sys.modules['__main__'] = main_mod
+                self._reset_main = True
+        if '__file__' in self.namespace:
+            self._previous_filename = self.namespace['__file__']
+        self.namespace['__file__'] = self.filename
+        return self.namespace
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        """
+        Reset the namespace.
+        """
+        if not self.current_namespace:
+            get_ipython().user_ns.update(self.namespace)
+        if self._previous_filename:
+            self.namespace['__file__'] = self._previous_filename
+        elif '__file__' in self.namespace:
+            # Avoid error when running `%reset -f` programmatically
+            # See issue spyder-ide/spyder-kernels#91
+            self.namespace.pop('__file__')
+        if self._previous_main:
+            sys.modules['__main__'] = self._previous_main
+        elif '__main__' in sys.modules and self._reset_main:
+            del sys.modules['__main__']
--- a/spyder_kernels/customize/spydercustomize.py
+++ b/spyder_kernels/customize/spydercustomize.py
@@ -907,6 +907,8 @@ class FileNamespace():
         """
         Reset namespace['__file__']
         """
+        if not self.current_namespace:
+            get_ipython().user_ns.update(self.namespace)
         if self._previous_filename:
             self.namespace['__file__'] = self._previous_filename
         elif '__file__' in self.namespace:
@@ -985,7 +987,6 @@ def runfile(filename=None, args=None, wdir=None, namespace=None,
                 # tb offset is 1 because we wrap execfile
                 ipython_shell.showtraceback(tb_offset=2)
 
-        ipython_shell.user_ns.update(namespace)
         clear_post_mortem()
         sys.argv = ['']
 
--- a/spyder_kernels/customize/spydercustomize.py
+++ b/spyder_kernels/customize/spydercustomize.py
@@ -29,6 +29,7 @@ from IPython.core.getipython import get_ipython
 
 from spyder_kernels.py3compat import TimeoutError
 from spyder_kernels.comms import CommError
+from spyder_kernels.customize.namespace_manager import NamespaceManager
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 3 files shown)


impact27:
I don't think these errors are related to my work:
```error
_________ ERROR collecting spyder_kernels/utils/tests/test_nsview.py __________
spyder_kernels\utils\tests\test_nsview.py:38: in <module>
    PANEL = pd.Panel({0: pd.DataFrame([1,2]), 1:pd.DataFrame([3,4])})
E   TypeError: object() takes no parameters
```

ccordoba12:
I think this one can be closed now that you merged it with #112, right?

impact27:
This can be merged after #112

ccordoba12:
@impact27, please clean the git history of this PR.

impact27:
The runfile function has this line: `ipython_shell.user_ns.update(namespace)`. Is that something that should also be run with runcell?

ccordoba12:
> The runfile function has this line: ipython_shell.user_ns.update(namespace). Is that something that should also be run with runcell?

@bcolsen?

bcolsen:
>The runfile function has this line: `ipython_shell.user_ns.update(namespace)`. Is that something that should also be run with runcell?

That's only necessary to deal with the merging the new namespace that the file was run in with the current console namespace. Runcell is always in the current console namespace.

With the new namespace class I think that `ipython_shell.user_ns.update(namespace)` would be best in the `__exit__` method there. It only needs to be run if `current_namespace` is `False`.

bcolsen:
What namespace is `debugfile` run in currently?

spyder_kernels/customize/spydercustomize.py
```diff
@@ -1007,26 +1052,25 @@ def runcell(cellname, filename=None):
...
-    # Restore __file__
-    if namespace_has_file:
-        namespace['__file__'] = namespace_filename
-    else:
-        namespace.pop('__file__')
+    with FileNamespace(filename, current_namespace=True) as namespace:
+        try:
+            if PY2:
+                filename = maybe_encode(filename)
+            exec(compile(cell_code, filename, 'exec'), namespace)
```
bcolsen: If we don't use IPython's `run_cell`, does that mean that IPython magics still work?

impact27: Do we expect them to work? I mean if I run a file they don't work, wo why should they work when running a cell?

impact27: In my opinion, cells are just file fragments

bcolsen: It would be nice. We are trying to make it so you can run a file as cells like a Jupyter notebook.

impact27:
So should runfile also accept magic?

spyder_kernels/customize/spydercustomize.py
bcolsen: Cell magic current works for cells in spyder3

bcolsen:
> So should runfile also accept magic?

The IPython `%run` magic only allows magics in .ipy and .ipynb they just run all the cells in a notebook file or the whole file in the case of an .ipy file. We could support .ipy file with spyder style cells or try to upstream that idea to IPython. We could also just have a button in spyder to run all the cells. 

https://github.com/ipython/ipython/blob/8971c8e5f31d20768609e20f60870d627eb33bd0/IPython/core/interactiveshell.py#L2741

It looks like they also can't get the file name for the trace backs but give a little warning at the end saying file they were running when the errors happened.

If we have to make a choice between having magics(I think IPython `run_cell` also does some async stuff and module reloads) or having filenames in tracebacks, I would pick magics and the stability of IPython `run_cell`(my opinion). 

In general IPython functions get a lot of testing and we should use then when we can. I know that this is working around problem with the debugger, but the closer we stay to upstream then the lighter the bug fixing load there is here. There are only so many corner cases we can test.

impact27:
The main problem I see is that if the name of the file is not set properly, you can not use breakpoints with Pdb. An alternative would be to parse the cell for magic and call IPython explicitly.

spyder_kernels/customize/spydercustomize.py
bcolsen: It looks like vscode can runcells with magics

bcolsen:
> An alternative would be to parse the cell for magic and call IPython explicitly.

IPython has `.run_cell_magic` and  `.run_line_magic` so that might work

impact27:
`ipython_shell.run_cell` was removed in #112, so I am not sure this is in scope of this PR. I think this should be discussed in a separate issue.

spyder_kernels/customize/spydercustomize.py
```diff
@@ -867,8 +867,62 @@ def get_debugger(filename):
...
+        Reset namespace['__file__']
+        """
+        if not self.current_namespace:
+            get_ipython().user_ns.update(self.namespace)
+        if self._previous_filename:
+            self.namespace['__file__'] = self._previous_filename
+        elif '__file__' in self.namespace:
+            # Avoid error when running `%reset -f` programmatically
+            # See issue spyder-ide/spyder-kernels#91
+            self.namespace.pop('__file__')
```
ccordoba12: Please enclose this in a try/except KeyError (as before) to avoid the issue referenced above this line.

spyder_kernels/customize/spydercustomize.py
```diff
@@ -867,8 +867,62 @@ def get_debugger(filename):
     return debugger, filename
 
 
+class FileNamespace():
```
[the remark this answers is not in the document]
ccordoba12: Also, please make this class inherit from `object` to be a new class in Python 2.

spyder_kernels/customize/spydercustomize.py
```diff
@@ -898,52 +952,43 @@ def runfile(filename=None, args=None, wdir=None, namespace=None,
...
+                ipython_shell.run_cell_magic('cython', '', f.read())
+        else:
+            try:
+                execfile(filename, namespace)
+            except SystemExit as status:
+                # ignore exit(0)
+                if status.code:
+                    ipython_shell.showtraceback(exception_only=True)
+            except bdb.BdbQuit:
+                if not is_pdb:
```
ccordoba12: Please add a comment above this line to describe what you're doing here.

spyder_kernels/customize/spydercustomize.py
```diff
@@ -898,52 +952,43 @@ def runfile(filename=None, args=None, wdir=None, namespace=None,
...
+                    ipython_shell.showtraceback(tb_offset=2)
+            except BaseException:
+                # tb offset is 1 because we wrap execfile
```
ccordoba12: Could you better describe here what part of the traceback we're discarding here?

spyder_kernels/customize/spydercustomize.py
```diff
@@ -1007,26 +1053,25 @@ def runcell(cellname, filename=None):
...
+        try:
+            if PY2:
+                filename = maybe_encode(filename)
+            exec(compile(cell_code, filename, 'exec'), namespace)
+        except SystemExit as status:
+            # ignore exit(0)
+            if status.code:
+                ipython_shell.showtraceback(exception_only=True)
+        except bdb.BdbQuit:
+            if not is_pdb:
```
ccordoba12: Please add the same comments here as above.

spyder_kernels/customize/spydercustomize.py
ccordoba12: I don't think we should allow magics in Python files because they're not valid code.

spyder_kernels/customize/spydercustomize.py
impact27: Isn't `if '__file__' in self.namespace` enough? I thought the `KeyError` was to catch that.

spyder_kernels/customize/spydercustomize.py
impact27: Sure, we discard the call to `execfile` and the call to `exec`

spyder_kernels/customize/spydercustomize.py
ccordoba12: `%reset -f` removes it after `runfile` is executed, so this could still happen (I haven't tested it with this PR though). Also, since you left the comment above, I thought it still happened.

spyder_kernels/customize/spydercustomize.py
ccordoba12: Ok, then please that explanation instead of the comment in this line.

spyder_kernels/customize/spydercustomize.py
impact27: I left the comment to explain the `if '__file__' in self.namespace`. I tried spyder-ide/spyder-kernels#91 and it works as expected

spyder_kernels/customize/namespace_manager.py
```diff
@@ -0,0 +1,72 @@
...
+        """
+        Reset the namespace.
+        """
+        if not self.current_namespace:
+            get_ipython().user_ns.update(self.namespace)
+        if self._previous_filename:
+            self.namespace['__file__'] = self._previous_filename
+        elif '__file__' in self.namespace:
+            # Avoid error when running `%reset -f` programmatically
+            # See issue spyder-ide/spyder-kernels#91
```
ccordoba12: The two comments above no longer apply after your changes in this PR, so please remove them.

PR #136, opened 2019-08-05
PR: Ask the frontend to save files before running them

impact27:
Give a chance to the frontend to save a file before running it.

This is the companion of spyder-ide/spyder#9966.

### DIFF
--- a/spyder_kernels/comms/commbase.py
+++ b/spyder_kernels/comms/commbase.py
@@ -455,7 +455,7 @@ class CommBase(object):
         """
         Handle an error that was raised on the other side asyncronously.
         """
-        error_wrapper.raise_error()
+        error_wrapper.print_error()
 
     def _sync_error(self, error_wrapper):
         """
--- a/spyder_kernels/customize/spydercustomize.py
+++ b/spyder_kernels/customize/spydercustomize.py
@@ -841,7 +841,7 @@ def runfile(filename, args=None, wdir=None, namespace=None, post_mortem=False):
     """
     try:
         # Save the open files
-        _frontend_request().savefiles()
+        _frontend_request().save_files()
     except (CommError, TimeoutError):
         logger.debug("Could not send save files before executing.")
 
--- a/spyder_kernels/customize/spydercustomize.py
+++ b/spyder_kernels/customize/spydercustomize.py
@@ -839,6 +839,11 @@ def runfile(filename, args=None, wdir=None, namespace=None, post_mortem=False):
     wdir: working directory
     post_mortem: boolean, whether to enter post-mortem mode on error
     """
+    try:
+        # Save the open files
+        _frontend_request().savefiles()
+    except (CommError, TimeoutError):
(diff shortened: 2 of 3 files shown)


ccordoba12:
@impact27, please remove all merges in this branch and leave a single one with master.

spyder_kernels/comms/commbase.py
```diff
@@ -455,7 +455,7 @@ def _async_error(self, error_wrapper):
         """
         Handle an error that was raised on the other side asyncronously.
         """
-        error_wrapper.raise_error()
+        error_wrapper.print_error()
```
ccordoba12: Why this change?

impact27: I don't want to raise the error for async call, because if the call is async, the caller will not be expecting an exception. This is only a problem on the kernel side as the frontend reimplements the function.

impact27: For example, if I make a syncronous call, I don't want to have an error for an unrelated async call being raised.

impact27: Thinking about it, I have the following question:
The kernel makes a call that fails on the frontend side. Where should the error go?
 - If the error is syncronous, it should go back to the kernel side, because the calling function might want to handle the error.
 - If the call is asyncronous, the kernel side should be notified as well, so the callback for this call can be discarded.

But should we send back async error to the frontend to be displayed in the error window?

impact27: I can do that easily. Should I open a new PR or attach this to this one?

ccordoba12: > But should we send back async error to the frontend to be displayed in the error window?

Yes, I think that's important to do it.

> Should I open a new PR or attach this to this one?

Please do it here (this PR is already very simple).

spyder_kernels/customize/spydercustomize.py
```diff
@@ -839,6 +839,11 @@ def runfile(filename, args=None, wdir=None, namespace=None, post_mortem=False):
     wdir: working directory
     post_mortem: boolean, whether to enter post-mortem mode on error
     """
+    try:
+        # Save the open files
+        _frontend_request().savefiles()
```
ccordoba12: I think this should contain `filename` as part of its arguments.

ccordoba12: I mean, because the editor needs to verify if the file is open there before trying to save it, right?

impact27: Well for now the frontend saves all the files when you run a script (eg. by pressing the run button). This makes sense because one file can call another. This just uses this already implemented mechanism. I can add the file name as an argument but it won't be used. 
For ref:
https://github.com/spyder-ide/spyder/blob/82cad82ef1fa454fa0703ce6db90609ab31784dd/spyder/plugins/editor/plugin.py#L2330

impact27: And you can run a file that is not open.

PR #137, opened 2019-08-20
PR: Add a way to change foreground color of Sympy repr's

juanis2112:
Added method for setting the colors of the sympy equations depending on the background color.

Screen Shot 2019-08-27 at 10 25 26 AM

Issue Spyder-ide/spyder#3798

### DIFF
--- a/spyder_kernels/console/kernel.py
+++ b/spyder_kernels/console/kernel.py
@@ -55,7 +55,7 @@ class SpyderKernel(IPythonKernel):
             'get_namespace_view': self.get_namespace_view,
             'set_namespace_view_settings': self.set_namespace_view_settings,
             'get_var_properties': self.get_var_properties,
-            'set_equations_color': self.set_equations_color
+            'set_sympy_forecolor': self.set_sympy_forecolor
             }
         for call_id in handlers:
             self.frontend_comm.register_call_handler(
@@ -486,6 +486,18 @@ class SpyderKernel(IPythonKernel):
         if self._mpl_backend_error is not None:
             print(self._mpl_backend_error)  # spyder: test-skip
 
+    def set_sympy_forecolor(self, background_color='dark'):
+        """Set SymPy forecolor depending on console background."""
+        if os.environ.get('SPY_SYMPY_O') == 'True':
+            try:
+                from sympy import init_printing
+                if background_color == 'dark':
+                    init_printing(forecolor='White')
+                elif background_color == 'light':
+                    init_printing(forecolor='Black')
+            except Exception:
+                pass
+
     # --- Others
     def _load_autoreload_magic(self):
         """Load %autoreload magic."""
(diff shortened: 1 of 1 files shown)


spyder_kernels/console/kernel.py
```diff
@@ -507,3 +508,12 @@ def _load_wurlitzer(self):
                 get_ipython().run_line_magic('reload_ext', 'wurlitzer')
             except Exception:
                 pass
+    def set_equations_color(self, background_color):
```
goanpeca: @ccordoba12 maybe we should define a default?

Also I think we should use, light/dark (I saw the example @juanis2112 used was white vs dark?

spyder_kernels/console/kernel.py
```diff
@@ -507,3 +508,12 @@ def _load_wurlitzer(self):
...
+        try:
+            from sympy import init_printing
+            if background_color=="dark":
```
goanpeca:
```suggestion
            if background_color == 'dark':
```

spyder_kernels/console/kernel.py
```diff
@@ -507,3 +508,12 @@ def _load_wurlitzer(self):
...
+                init_printing(forecolor='White')
+            else:
```
goanpeca:
```suggestion
            elif background_color == 'light':
```

spyder_kernels/console/kernel.py
```diff
@@ -507,3 +508,12 @@ def _load_wurlitzer(self):
...
+                init_printing(forecolor='Black')
+        except:
+            pass
```
goanpeca: Should we emit some warning that the method "failed" or something @ccordoba12 ?

juanis2112:
@goanpeca I fixed the minor issues but let's wait for @ccordoba12  regarding the warning message.

spyder_kernels/console/kernel.py
ccordoba12: I don't want to report an error if Sympy is not installed, but excepting on ImportError sometimes picks up other nasty errors with the package dependencies.

So what do you propose?

goanpeca: We could check if it is not an import error (use except Exception) and if it is not emit some warning with whatever the error was?

ccordoba12:
Thanks a lot @juanis2112!! This looks pretty good to me now.

Let's wait for @dalthviz to implement his part on the Spyder side to see if we need to add or modify something else here.

spyder_kernels/console/kernel.py
```diff
@@ -485,6 +486,18 @@ def show_mpl_backend_errors(self):
         if self._mpl_backend_error is not None:
             print(self._mpl_backend_error)  # spyder: test-skip
 
+    def set_sympy_forecolor(self, background_color='dark'):
+        """Set SymPy forecolor depending on console background."""
+        if os.environ.get('SPY_SYMPY_O') == 'True':
+            try:
+                from sympy import init_printing
+                if background_color == 'dark':
+                    init_printing(forecolor='White')
```
dalthviz: @ccordoba12 @juanis2112 after testing this with spyder-ide/spyder#10114 the `init_printing` call also needs the [`ip` parameter](https://github.com/sympy/sympy/blob/0ddeeea520a273bb31e672a549529caa2ed91d9c/sympy/interactive/printing.py#L369) (the IPython instance needs to be passed).

For that @juanis2112 please:
* Add the import for `get_ipython` (`from IPython.core.getipython import get_ipython`) below the one for `Sympy`,
* Then use it in the call of `init_printing`, it should look something like `init_printing(forecolor='White', ip=get_ipython())`  (this in each call made).

goanpeca:
@dalthviz is this good to go?

dalthviz:
@goanpeca I will check :+1:

PR #138, opened 2019-08-21
PR: Use IPython safe_execfile to execute files and clean tracebacks

bcolsen:
This change uses IPython's `safe_execfile` function to execute files in `runfile`. 

This `safe_execfile` catches and prints the file error allowing execution of the `runfile` function to continue after an error in the file. This allows us to still load the namespace of the file into the global namespace after an error. 

It also has the nice side effect of giving clean error tracebacks. @impact27, will #134 be necessary after this change? 

Before:
```python traceback

  File "<ipython-input-1-664d90654f60>", line 1, in <module>
    runfile('/home/bcolsen/Documents/python/jv_stat_bing/plot_data_set.py', wdir='/home/bcolsen/Documents/python/jv_stat_bing')

  File "/home/bcolsen/anaconda3/lib/python3.6/site-packages/spyder_kernels-1.5.0.dev0-py3.6.egg/spyder_kernels/customize/spydercustomize.py", line 890, in runfile
    execfile(filename, namespace)

  File "/home/bcolsen/anaconda3/lib/python3.6/site-packages/spyder_kernels-1.5.0.dev0-py3.6.egg/spyder_kernels/customize/spydercustomize.py", line 116, in execfile
    exec(compile(f.read(), filename, 'exec'), namespace)

  File "/home/bcolsen/Documents/python/jv_stat_bing/plot_data_set.py", line 500, in <module>
    yo

NameError: name 'yo' is not defined
```

After
```python traceback

  File "/home/bcolsen/Documents/python/jv_stat_bing/plot_data_set.py", line 500, in <module>
    yo

NameError: name 'yo' is not defined
```

The default to cleanly exit without an error with `sys.exit()` making the work flow in https://github.com/spyder-ide/spyder/issues/8100 more clean. 

This might affect other areas as well.

### DIFF
--- a/spyder_kernels/console/tests/test_console_kernel.py
+++ b/spyder_kernels/console/tests/test_console_kernel.py
@@ -368,7 +368,7 @@ def test_runfile(tmpdir):
         client.get_shell_msg(block=True, timeout=TIMEOUT)
 
         # Write defined variable code to a file
-        code = u"result = 'hello world'"
+        code = u"result = 'hello world'; error # make an error"
         d = tmpdir.join("defined-test.py")
         d.write(code)
 
@@ -382,7 +382,7 @@ def test_runfile(tmpdir):
         u = tmpdir.join("undefined-test.py")
         u.write(code)
 
-        # Run code file `d` to define `result`
+        # Run code file `d` to define `result` even after error
         client.execute("runfile(r'{}', current_namespace=False)"
                        .format(to_text_string(d)))
         client.get_shell_msg(block=True, timeout=TIMEOUT)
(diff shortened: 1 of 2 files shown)


impact27:
I looked into using `safe_execfile`, but it causes problem for debugging. Specifically, in `spydercustomize.py`:
```
@monkeypatch_method(pdb.Pdb, 'Pdb')
def interaction(self, frame, traceback):
    if frame is not None and "spydercustomize.py" in frame.f_code.co_filename:
        self.run('exit')
    ...
```
If you use `safe_execfile`, `self.run('exit')` would not be correctly called, so you need to update this function as well.

 #134 also fixes this prooblem for `runcell`, and fixes `__file__` and `__main__` issues.

bcolsen:
I see. It actually looks like #134 might solve the problem I was trying to solve here with the name space not updating as well I will test it.

bcolsen:
Haha I should have checked #134 first. The change there fix the problems I'm having with running in a new namespace.

@ccordoba12 I'm closing this. I think everything will be fine here as long as #134 gets merged before the next spyder-kernels release.

@impact27 Can you add the change I made to the runfile test  into #134?

PR #141, opened 2019-08-27
[PR] Fix value of sys.path[0] when executing file

jackvreeken:
According to the Python specification/documentation, the value of
sys.path[0] is that of the directory containing the file that was
executed. This behavior was not properly replicated in runfile().

I have no idea if I implemented it correctly like this, as there are a whole bunch of routines that seem to execute files. Seeing as `runfile()` was the one that changed `sys.argv`, I felt it appropriate to also set the correct value for `sys.path[0]` here.

### DIFF
--- a/spyder_kernels/console/tests/test_console_kernel.py
+++ b/spyder_kernels/console/tests/test_console_kernel.py
@@ -416,6 +416,42 @@ def test_runfile(tmpdir):
         content = msg['content']
         assert content['found']
 
+        # Check if file environment variables like __file__ and sys.path are set correctly
+        client.execute("%reset -f")
+        client.get_shell_msg(block=True, timeout=TIMEOUT)
+
+        code = dedent(u"""
+        import os
+        import sys
+
+        try:
+            env_file = __file__
+        except NameError:
+            env_file = None
+
+        try:
+            env_sys_path_0 = sys.path[0]
+        except NameError:
+            env_sys_path_0 = None
+
+        if (env_file is not None
+                and env_sys_path_0 is not None
+                and (os.path.dirname(os.path.abspath(__file__)) ==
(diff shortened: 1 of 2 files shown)


impact27:
https://docs.python.org/3.7/library/sys.html#sys.path

> As initialized upon program startup, the first item of this list, path[0], is the directory containing the script that was used to invoke the Python interpreter. If the script directory is not available (e.g. if the interpreter is invoked interactively or if the script is read from standard input), path[0] is the empty string, which directs Python to search modules in the current directory first. Notice that the script directory is inserted before the entries inserted as a result of PYTHONPATH.

This should probably go in `NamespaceManager`.

jackvreeken:
Alright, I moved it to NamespaceManager as suggested, and added a test.

One thing I'm not sure of is what happens (or what to do) when the file in question starts playing with sys.path. Should we keep those changes around?

impact27:
I think the user would expect the changes made to sys.path to be persistant. This is the same with cwd. The cleanup should probably be: remove the first item if it is what we added.

jackvreeken:
I think this should about cover it then.

impact27:
IPython 7.2+ doesn't behave like python for `sys.path[0]`, so I don't think we should change the behaviour here.

From `IPython/core/shellapp.py::InteractiveShellApp.init_path`:
```
        Unlike Python's default, we insert before the first `site-packages`
        or `dist-packages` directory,
        so that it is after the standard library.

        .. versionchanged:: 7.2
            Try to insert after the standard library, instead of first.
```

ccordoba12:
> IPython 7.2+ doesn't behave like python for sys.path[0], so I don't think we should change the behaviour here.

Agreed. We basically embed IPython in Spyder, so we should not change its behavior to make it the same as the default Python one.

Therefore I'm closing this one.

PR #152, opened 2019-09-15
PR: Add a way to ignore installed Python libraries while debugging

impact27:
Add an option to ignore libraries while debugging.
Solve part of https://github.com/spyder-ide/spyder/issues/3232:

>  2. "step into my code" (from PyCharm) "helps skip stepping into library sources and keep focused on your own code", very useful to avoid enter into functions from python and other modules.

see https://github.com/spyder-ide/spyder/pull/10199


bcolsen:
This is a cool idea. Wouldn't be possible to get the python path from sys.path and not step into files in 'lib/python*' from there.

impact27:
sure, should that include the `up` and `down` commands?

impact27:
@bcolsen Your idea was better so I changed the PR. I am not using `sys.path` as it might include user code that is install with `pip -e`

bcolsen:
> sure, should that include the up and down commands?

I don't know. Not stepping into the a lib is probably what most people would want step to do. But if the only way 'down' is into a library then I guess we let them go there.

> I am not using sys.path as it might include user code that is installed in editable mode (pip install -e).

Yeah `site` is seems like the better idea here.

spyder_kernels/customize/spydercustomize.py
```diff
@@ -422,6 +426,32 @@ def interaction(self, frame, traceback):
...
+            return False
+        filename = frame.f_code.co_filename
+        if filename.startswith('<'):
+            # This is not a file
+            return True
+        if self.pdb_ignore_lib and self.path_is_lib(filename):
+            return False
+        return True
+
+    def path_is_lib(self, path):
```
ccordoba12: We have a method in the `UserModuleReloader` class that does exactly this. So please unify these two code in a single function an use it in both here and there.

impact27: Done

ccordoba12:
@impact27, I improved the code a bit for readability and to make it more Pythonic. I also fixed one of the tests you changed so we can verify that the usage of `path_is_library` is correct when called with the UMR.

I think this is ready to be merged, but please take a look at it to see if my changes didn't alter your previous work.

impact27:
Looks good

spyder_kernels/console/kernel.py
```diff
@@ -114,6 +115,13 @@ def set_pdb_echo_code(self, state):
         """
         self._pdb_print_code = state
 
+    def set_pdb_ignore_lib(self, state):
+        """
+        Handle a message from the frontend
```
goanpeca: Please add a period :-p

PR #153, opened 2019-09-16
PR: Disable echo

impact27:
Add an option to disable code echo


spyder_kernels/customize/spydercustomize.py
```diff
@@ -418,7 +418,8 @@ def interaction(self, frame, traceback):
             self.setup(frame, traceback)
             if self.send_initial_notification:
                 self.notify_spyder(frame)
-            self.print_stack_entry(self.stack[self.curindex])
+            if get_ipython().kernel._pdb_print_code:
```
impact27: add `_pdb_print_code` to pdb settings

ccordoba12: We decided that this was not necessary because Pdb settings are used to initialize the Pdb instance, but this value can change on every evaluation.

PR #155, opened 2019-09-19
Process pending messages and matplotlib events while waiting for input.

impact27:
See PR https://github.com/ipython/ipykernel/pull/438

Adding this PR to ipykernel looks complicated as this would cause messages being potentially processed out of order. Here, we control the frontend (In most cases), so processing the messages should be safer.

The potential problems are that a message like 'execute_request' might be recieved by the kernel while debugging or waiting for a general input. The Spyder frontend doesn't do that (as far as I know) but another frontend might.

`spyder-kernels` is already exposed to this problem as calling `do_one_iteration` is how the comms can do blocking calls. It is also used in my 'complete during debugging' PR.

I think most users of `spyder-kernels` will use it with spyder. So I would rather support seeing plots while debugging than not risk a bug popping up if someone uses another frontend with spyder kernels.

A better implementation could be done but probably not before spyder4:

Pdb should make some kind of special request (This could be generalized to other command line system). What I would see is:
 - The input request is sent to a single client (If I am not mistaken). Therefore only messages from this client should be processed while waiting for an input. (Are messages sent from different clients supposed to be processed in order, or just messages from a single client?)
 - The input request should contain some kind of field that notifies the frontend that this is a special input request, so the frontend could do act accordingly. For example, it would know it is ok to send messages as they could be processed. (Such as comm, complete_request, ...)
 - The kernel would process only messages that the frontend sends with this special value in the header. If the frontend sends a non-special message the processing would stop.

### DIFF
--- a/spyder_kernels/console/kernel.py
+++ b/spyder_kernels/console/kernel.py
@@ -529,13 +529,11 @@ class SpyderKernel(IPythonKernel):
                 pass
 
     def _input_request(self, prompt, ident, parent, password=False):
-        """Send an input request to the frontend and wait for the reply."""
-        # matplotlib needs to be imported after app.launch_new_instance()
-        try:
-            from matplotlib._pylab_helpers import Gcf
-        except ImportError:
-            Gcf = None
+        """Send an input request to the frontend and wait for the reply.
 
+        This function can be removed if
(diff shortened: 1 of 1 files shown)


impact27:
Closing in favor of https://github.com/ipython/ipykernel/pull/438

PR #158, opened 2019-09-26
PR: Allow comms to works in threads

impact27:
This create a asyncio loop to allow calling `do_one_iteration` 

Fixes spyder-ide/spyder#8801


impact27:
This is useful e.g. for debugging threads:
```
from threading import Thread
import time


class MyThread(Thread):
    def __init__(self):
        Thread.__init__(self)
        self.start()

    def run(self):
        import pdb
        pdb.set_trace()
        for idx in range(10):
            time.sleep(1)
            print(idx)


MyThread().join()
```

impact27:
Superseded by #170

PR #159, opened 2019-09-29
PR: handle closed files in the frontend 

impact27:
Adds to https://github.com/spyder-ide/spyder-kernels/pull/143 to move the handling of a closed file


impact27:
This is not a good idea to remove this fallback

PR #163, opened 2019-10-04
PR: Fix tests that use setup_kernel

impact27:
Repeat tests that uses setup_kernel


impact27:
The AppVeyor test really doesn't want to pass.

PR #169, opened 2019-10-17
PR: Create a Comm socket

impact27:
Calling `do_one_iteration` is not a good idea as it might process messages out of order.

Here, a new socket is created. It is a one way socket designed to skip the message queue, and runs in a thread. This means that:
 - Only messages that are explicitely sent to this socket are processed out of order, instead of all messages as it is the case now.
 - Any thread can make a blocking call to the frontend without needing to start an event loop.
 - We can set breakpoints while executing!

### DIFF
--- a/spyder_kernels/comms/frontendcomm.py
+++ b/spyder_kernels/comms/frontendcomm.py
@@ -69,6 +69,7 @@ class FrontendComm(CommBase):
             # Create an event loop to handle some messages
             ioloop.IOLoop().initialize()
         while True:
+            out_stream = self.kernel.shell_streams[0]
             try:
                 ident, msg = self.kernel.session.recv(self.comm_socket, 0)
             except Exception:
@@ -81,13 +82,15 @@ class FrontendComm(CommBase):
             else:
                 try:
                     # shell_streams[0] handles replies
-                    handler(self.kernel.shell_streams[0], ident, msg)
+                    handler(out_stream, ident, msg)
                 except Exception:
                     self.kernel.log.error("Exception in message handler:",
                                           exc_info=True)
 
             sys.stdout.flush()
             sys.stderr.flush()
+            # flush to ensure reply is sent
+            out_stream.flush(zmq.POLLOUT)
 
     def remote_call(self, comm_id=None, blocking=False, callback=None):
         """Get a handler for remote calls."""
--- a/spyder_kernels/comms/frontendcomm.py
+++ b/spyder_kernels/comms/frontendcomm.py
@@ -16,6 +16,7 @@ import sys
 from zmq.eventloop.zmqstream import ZMQStream
 import socket
 from jupyter_client.localinterfaces import localhost
+from tornado import ioloop
 
 from spyder_kernels.comms.commbase import CommBase
 from spyder_kernels.py3compat import TimeoutError, PY2
@@ -58,13 +59,15 @@ class FrontendComm(CommBase):
             # in certain rare circumstances
             # see ipython/ipykernel#270 and zeromq/libzmq#2892
             self.comm_socket.router_handover = 1
-        self.comm_stream = ZMQStream(self.comm_socket)
 
         self.comm_socket_thread = threading.Thread(target=self.poll_thread)
         self.comm_socket_thread.start()
 
     def poll_thread(self):
         """Recieve messages from comm socket"""
+        if not PY2:
+            # Create an event loop to handle some messages
+            ioloop.IOLoop().initialize()
         while True:
             try:
                 ident, msg = self.kernel.session.recv(self.comm_socket, 0)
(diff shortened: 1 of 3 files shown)


spyder_kernels/comms/frontendcomm.py
```diff
@@ -27,8 +43,81 @@ def __init__(self, kernel):
...
+                    self.comm_socket_thread.join()
+                    super_close()
+
+                self.kernel.parent.close = close
+
+    def poll_thread(self):
+        """Recieve messages from comm socket."""
         if not PY2:
-            self._main_thread_id = threading.get_ident()
+            # Create an event loop to handle some messages
```
ccordoba12: Let's remove `some` in this comment, because I guess this event loop handles all messages, right?

spyder_kernels/comms/frontendcomm.py
```diff
@@ -27,8 +43,81 @@ def __init__(self, kernel):
...
+            ioloop.IOLoop().initialize()
+        while not self.comm_thread_close.is_set():
+            self.poll_one()
+
+    def poll_one(self):
+        """Recieve one message from comm socket."""
```
ccordoba12: Recieve -> Receive

spyder_kernels/comms/frontendcomm.py
```diff
@@ -27,8 +43,81 @@ def __init__(self, kernel):
...
+        if msg_type == 'shutdown_request':
+            self.comm_thread_close.set()
+            return
+
+        handler = self.kernel.shell_handlers.get(msg_type, None)
+        if handler is None:
+            self.kernel.log.warning("Unknown message type: %r", msg_type)
+        else:
+            try:
+                # shell_streams[0] handles replies
```
ccordoba12: Please expand this comment because I don't understand very well why you're handling the comm messages in the `out_stream` (at least that's what it looks like by reading the code below).

spyder_kernels/customize/spydercustomize.py
```diff
@@ -313,6 +313,15 @@ def __init__(self, completekey='tab', stdin=None, stdout=None,
         self._pdb_breaking = False
 
     # --- Methods overriden by us
+    def set_continue(self):
+        """Stop only at breakpoints or when finished.
```
ccordoba12: When you're writing multi-line strings, our custom is to leave the initial triple quotes empty. In other words, please rewrite this as

```python
"""
Stop only at breakpoints or when finished.

...
"""
```

spyder_kernels/comms/frontendcomm.py
impact27: The `out_stream` is used by the handler to send replies. Usually the shell stream is used to both recieve and send messages. Here, we replace the "recieve" but not the "send", so we still need `out_stream` to send

spyder_kernels/comms/frontendcomm.py
impact27: It is only needed if `out_stream` is used by the handler. For now this is only used by the complete requests.

spyder_kernels/comms/frontendcomm.py
```diff
@@ -27,8 +43,81 @@ def __init__(self, kernel):
...
+            context = zmq.Context()
+            self.comm_socket = context.socket(zmq.ROUTER)
+            self.comm_socket.linger = 1000
+
+            self.comm_port = get_port()
+
+            self.comm_port = self.kernel.parent._bind_socket(
+                self.comm_socket, self.comm_port)
+            if hasattr(zmq, 'ROUTER_HANDOVER'):
+                # set router-handover to workaround zeromq reconnect problems
```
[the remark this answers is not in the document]
ccordoba12: This is unresolved yet.

spyder_kernels/comms/frontendcomm.py
```diff
@@ -27,8 +43,81 @@ def __init__(self, kernel):
...
+                # in certain rare circumstances
```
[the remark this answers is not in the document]
ccordoba12: Unresolved.

spyder_kernels/comms/frontendcomm.py
```diff
@@ -27,8 +43,81 @@ def __init__(self, kernel):
...
+                # see ipython/ipykernel#270 and zeromq/libzmq#2892
```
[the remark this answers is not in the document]
ccordoba12: Unresolved.

spyder_kernels/comms/frontendcomm.py
```diff
@@ -27,8 +43,81 @@ def __init__(self, kernel):
...
+                self.comm_socket.router_handover = 1
+
+            self.comm_thread_close = threading.Event()
+            self.comm_socket_thread = threading.Thread(target=self.poll_thread)
+            self.comm_socket_thread.start()
+
+            # patch parent.close
```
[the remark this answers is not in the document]
ccordoba12: Unresolved.

spyder_kernels/comms/frontendcomm.py
```diff
@@ -27,8 +43,81 @@ def __init__(self, kernel):
...
+            if not PY2:
+                super_close = self.kernel.parent.close
```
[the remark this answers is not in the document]
ccordoba12: Unresolved.

ccordoba12:
@impact27, there are still several comments that are not solved.

PR #171, opened 2019-10-19
PR: Publish Pdb stack frames to Spyder

impact27:
Enable getting and printing the current state of the console


impact27:
Why is there python 2.7 tests in master?

spyder_kernels/console/kernel.py
```diff
@@ -145,12 +148,74 @@ def disable_faulthandler(self):
...
 
+    def get_system_threads_id(self):
+        """Return the list of system threads id."""
+        ignore_threads = [
+            self.parent.poller,  # Parent poller
+            self.shell.history_manager.save_thread,  # history
+            self.parent.heartbeat,  # heartbeat
+            self.parent.iopub_thread.thread,  # iopub
+            gc.thread,  # ZMQ garbage collector thread
+            ]
```
ccordoba12, line 159:
```suggestion
        ]
```

spyder_kernels/console/kernel.py
```diff
@@ -145,12 +148,74 @@ def disable_faulthandler(self):
...
+        if hasattr(self.parent, "control_thread"):
+            ignore_threads.append(self.parent.control_thread)
```
ccordoba12, line 161: I think this is now always the case because we depend on jupyter-client > 7.

spyder_kernels/console/kernel.py
```diff
@@ -145,12 +148,74 @@ def disable_faulthandler(self):
...
+    def get_current_frames(self, ignore_internal_threads=True,
+                           capture_locals=False):
+        """Get the current frames."""
+        ignore_list = self.get_system_threads_id()
+        main_id = threading.main_thread().ident
+        frames = {}
+        thread_names = {thread.ident: thread.name
+                        for thread in threading.enumerate()}
+
+        for threadId, frame in sys._current_frames().items():
```
ccordoba12, line 194: Is it possible to change `threadId` for `thread_id` here?

That's because we prefer to use snake-cased variables instead of camel-cased ones. But if it's not possible, then it's fine.

spyder_kernels/console/shell.py
```diff
@@ -13,10 +13,14 @@
...
 import bdb
 import sys
+import traceback
 
 # Third-party imports
 from ipykernel.zmqshell import ZMQInteractiveShell
 
+# Local imports
+from spyder_kernels.utils.mpl import automatic_backend
+
```
ccordoba12, line 23: This import seems unnecessary. If that's the case, please remove it.

spyder_kernels/customize/spyderpdb.py
```diff
@@ -705,6 +705,26 @@ def notify_spyder(self):
 
         get_ipython().kernel.publish_pdb_state(step)
 
+        # Publish Pdb state so we can update the Variable Explorer
```
ccordoba12, line 708:
```suggestion
        # Publish Pdb stack so we can update the Variable Explorer
```

spyder_kernels/customize/spyderpdb.py
```diff
@@ -705,6 +705,26 @@ def notify_spyder(self):
...
+        if skip_hidden:
+            # Filter out the hidden frames
+            hidden = self.hidden_frames(self.stack)
+            pdb_stack = [f for f, h in zip(pdb_stack, hidden) if not h]
+            # Adjust the index
+            pdb_index -= sum(hidden[:pdb_index])
+
+        try:
+            frontend_request(blocking=False).set_pdb_stack(
+                pdb_stack, pdb_index)
```
ccordoba12, line 724: Do you need to do adjustments here after PR #381? Just asking because I'm not really sure.

ccordoba12 (commented):
Thanks @impact27 for your patience with this one. Besides my simple comments and suggestions, it looks good to me.

spyder_kernels/customize/spyderpdb.py
impact27: No I think PR #381 improves this one, but because we are adding to the end of the function everything is good

ccordoba12 (approved):

PR #175, opened 2019-10-29
PR: Add an option to execute IPython events in Pdb

impact27:
Add settings to trigger "pre_execute" and "post_execute" events when processing pdb code. This will for example show inline plots.

It depends on #152 as it uses the `pdb_settings` mechanism.


ccordoba12:
@impact27, this has a conflict with master. Please solve it so I can merge it.

PR #177, opened 2019-10-31
PR: Update ipykernel required version

impact27:
Fixes https://github.com/spyder-ide/spyder/issues/2902

### DIFF
--- a/requirements/posix.txt
+++ b/requirements/posix.txt
@@ -1,3 +1,3 @@
-ipykernel
+ipykernel>=5.1.3
 cloudpickle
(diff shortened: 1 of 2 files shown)


impact27:
`ipykernel` dropped python 2 so I am not sure how to require the correct version of ipykernel

ccordoba12:
> ipykernel dropped python 2 so I am not sure how to require the correct version of ipykernel

The fix you made was not backported to the ipykernel Python 2 branch, so what you did was the right thing to do.

PR #178, opened 2019-11-01
PR: Correctly set namespace while debugging

impact27:
`runfile` and `runcell` can be used while debugging to run code. However, the namespace is not properly set. This PR solves this problem.

### DIFF
--- a/spyder_kernels/customize/namespace_manager.py
+++ b/spyder_kernels/customize/namespace_manager.py
@@ -19,7 +19,7 @@ def _get_globals_locals():
         ns_locals = pdb.curframe_locals
         ns_globals = pdb.curframe.f_globals
     else:
-        ns_locals = {}
+        ns_locals = None
         ns_globals = get_ipython().user_ns
     return ns_globals, ns_locals
 
@@ -49,7 +49,7 @@ class NamespaceManager(object):
                  file_code=None):
         self.filename = filename
         self.ns_globals = namespace
-        self.ns_locals = {}
+        self.ns_locals = None
         self.current_namespace = current_namespace
         self._previous_filename = None
         self._previous_main = None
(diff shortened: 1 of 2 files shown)


spyder_kernels/customize/namespace_manager.py
```diff
@@ -12,10 +12,22 @@
 from spyder_kernels.py3compat import PY2
 
 
-def _get_globals():
-    """Return current namespace"""
-    ipython_shell = get_ipython()
-    return ipython_shell.user_ns
+def _get_globals_locals():
```
ccordoba12: We already have a method called `_get_current_namespace` in SpyderKernel, which is almost exactly as this one. I think it'd be better to use that instead.

spyder_kernels/customize/namespace_manager.py
```diff
@@ -12,10 +12,22 @@
...
+        pdb = get_ipython().kernel._pdb_obj
+        ns_locals = pdb.curframe_locals
+        ns_globals = pdb.curframe.f_globals
+    else:
+        ns_locals = None
+        ns_globals = get_ipython().user_ns
+    return ns_globals, ns_locals
+
+
+def _set_globals_locals(ns_globals, ns_locals):
```
ccordoba12: Why does this function need an `ns_locals` arg that it's not using?

spyder_kernels/customize/namespace_manager.py
```diff
@@ -29,7 +41,7 @@ class NamespaceManager(object):
     def __init__(self, filename, namespace=None, current_namespace=False,
                  file_code=None):
         self.filename = filename
-        self.namespace = namespace
+        self.namespace = namespace, None
```
ccordoba12: Why do you need to add `None` here? That's really not clear to me.

ccordoba12: It seems you're using `self.namespace` to refer to `globals` and `locals`. If that's the case, please use `self.ns_globals` and `self.ns_locals` instead.

spyder_kernels/customize/namespace_manager.py
impact27: I want to get both global and local separately as they are different arguments to `exec`. `_get_current_namespace` returns a mix of the two in a single dictionnary.

PR #180, opened 2019-11-03
PR: Cleanup spydercustomize

impact27:
This PR moves code around to remove SpyderPdb from spydercustomize.py and does a general cleanup of spydercustomize.py. No feature is added.


ccordoba12:
@impact27, please don't add more commits here. I added some to the nice refactoring you started here and I need to add some more, and then I'll merge it.

PR #188, opened 2019-11-28
PR: Print chained exceptions

impact27:
Fixes https://github.com/spyder-ide/spyder/issues/2943
Fixes https://github.com/spyder-ide/spyder/issues/1052

### DIFF
--- a/spyder_kernels/console/kernel.py
+++ b/spyder_kernels/console/kernel.py
@@ -81,7 +81,7 @@ class SpyderKernel(IPythonKernel):
             self, exc_tuple=None, filename=None, tb_offset=None,
             exception_only=False, running_compiled_code=False):
         """
-        Display the exception that just occurred, and any chined exception.
+        Display the exception that just occurred, and any chained exception.
 
         This is needed because we're using the plain style of showing
         tracebacks in qtconsole. And we're doing that to be able to parse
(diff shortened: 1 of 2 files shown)


spyder_kernels/console/kernel.py
```diff
@@ -73,6 +73,53 @@ def __init__(self, *args, **kwargs):
...
 
+        # Hack to solve spyder-ide/spyder/2943
+        self.shell_showtraceback = self.shell.showtraceback
+        self.shell.showtraceback = self.chained_showtraceback
+
+    def chained_showtraceback(
+            self, exc_tuple=None, filename=None, tb_offset=None,
+            exception_only=False, running_compiled_code=False):
+        """
+        Display the exception that just occurred, and any chined exception.
```
goanpeca: chined -> chained

ccordoba12:
@impact27, this is not the right way to do it. You simply need to remove this line

https://github.com/spyder-ide/spyder-kernels/blob/2c5b36cdb797b8aba77bc406ca96f5e079c4aaca/spyder_kernels/console/start.py#L103

But you also need to update the `TracebackLinksMixin` class in order for file strings in chained exceptions to be clickable and can be opened in our editor.

impact27:
The result would be different:
Regular:
Screenshot 2019-11-29 at 06 10 26
New:
Screenshot 2019-11-29 at 06 12 22

ccordoba12:
I'd really prefer that we switch to show tracebacks as it's done in a regular IPython session, to unify the behavior with the notebook and JupyterLab.

PR #189, opened 2019-11-29
PR: Patch PyQt to save created QApplication instances

impact27:
Fixes https://github.com/spyder-ide/spyder/issues/2970

### DIFF
--- a/spyder_kernels/customize/spydercustomize.py
+++ b/spyder_kernels/customize/spydercustomize.py
@@ -150,19 +150,19 @@ else:
 # Python sees all the QApplication as differnet python objects, while
 # Qt sees them as a singleton (There is only one Application!). Deleting one
 # QApplication causes all the other python instances to become broken.
+# See spyder-ide/spyder/issues/2970
 try:
     from PyQt5 import QtWidgets
 
-    original_5_QApplication = QtWidgets.QApplication
+    class SpyderQApplication(QtWidgets.QApplication):
+        def __init__(self, *args, **kwargs):
+            super(SpyderQApplication, self).__init__(*args, **kwargs)
+            # Add reference to avoid destruction
+            # This creates a Memory leak but avoids a Segmentation fault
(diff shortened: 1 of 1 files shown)


spyder_kernels/customize/spydercustomize.py
```diff
@@ -143,6 +143,47 @@ def _getfilesystemencoding_wrapper():
...
+    original_5_QApplication = QtWidgets.QApplication
+
+    def new_QApplication(*args, **kwargs):
+        instance = original_5_QApplication(*args, **kwargs)
+        if not hasattr(new_QApplication, 'wrapper_list'):
+            new_QApplication.wrapper_list = []
+        new_QApplication.wrapper_list.append(instance)
+        return instance
+
+    QtWidgets.QApplication = new_QApplication
```
dalthviz: Maybe this should be  `QtWidgets.QApplication.QApplication = new_QApplication` ?

spyder_kernels/customize/spydercustomize.py
```diff
@@ -143,6 +143,47 @@ def _getfilesystemencoding_wrapper():
...
+# Patch PyQt4 and PyQt5
+# =============================================================================
+# This saves the QApplication instances so that python doesn't destroy them.
+# Python sees all the QApplication as differnet python objects, while
+# Qt sees them as a singleton (There is only one Application!). Deleting one
+# QApplication causes all the other python instances to become broken.
+try:
+    from PyQt5 import QtWidgets
+
+    original_5_QApplication = QtWidgets.QApplication
```
dalthviz: Maybe a subclass of `QApplication` could be better here? Something like:

```python
    class SpyderQApplication(QtWidgets.QApplication):
            instance = original_5_QApplication.instance()
            if instance is None:
                super(SpyderQApplication, self).__init__(*args, **kwargs)

    QtWidgets.QApplication = SpyderQApplication
```

PR #195, opened 2020-01-08
PR: Give access to the running namespace when refreshing variables

impact27:
Add informations about the running namespace so the variable explorer can display informations about running variables when refreshing.


spyder_kernels/customize/namespace_manager.py
```diff
@@ -78,6 +78,10 @@ def __enter__(self):
         if '__file__' in self.ns_globals:
             self._previous_filename = self.ns_globals['__file__']
         self.ns_globals['__file__'] = self.filename
+
+        # Add the current namedspace to the kernel so it can access
```
ccordoba12: Please change this to

    # Add the current namespace to the kernel so it can access it

impact27:
> @impact27, I don't understand the purpose of this PR. Could you explain it a bit more?

When you refresh the variables in the variables explorer, this gives access to the variables even if runfile creates a new namespace.

ccordoba12:
> When you refresh the variables in the variables explorer, this gives access to the variables even if runfile creates a new namespace.

Ok, that's pretty cool!

Please solve the merge conflict so I can merge this one for 1.9.0. Thanks!

PR #198, opened 2020-01-21
PR: Add method to check dependencies for special consoles

dalthviz:
Needed for spyder-ide/spyder#11357

### DIFF
--- a/spyder_kernels/console/kernel.py
+++ b/spyder_kernels/console/kernel.py
@@ -19,6 +19,7 @@ from ipykernel.ipkernel import IPythonKernel
 
 # Local imports
 from spyder_kernels.comms.frontendcomm import FrontendComm
+from spyder_kernels.py3compat import to_text_string
 
 
 # Excluded variables from the Variable Explorer (i.e. they are not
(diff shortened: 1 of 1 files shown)


spyder_kernels/console/kernel.py
```diff
@@ -372,6 +374,21 @@ def close_all_mpl_figures(self):
...
+        """
+        try:
+            if os.environ.get('SPY_AUTOLOAD_PYLAB_O') == 'True':
+                import matplotlib
+            elif os.environ.get('SPY_SYMPY_O') == 'True':
+                import sympy
+            elif os.environ.get('SPY_RUN_CYTHON') == 'True':
+                import cython
+        except ImportError as e:
+            return to_text_string(e)
```
ccordoba12: Let's change this to

```python
except Exception as e:
    if os.environ.get('SPY_AUTOLOAD_PYLAB_O') == 'True':
        return u'matplotlib'
    elif os.environ.get('SPY_SYMPY_O') == 'True':
        return u'sympy'
    elif os.environ.get('SPY_RUN_CYTHON') == 'True':
        return 'cython'
```

On the Spyder side we'll use those names to build a generic error message.

spyder_kernels/console/kernel.py
```diff
@@ -19,6 +19,7 @@
 
 # Local imports
 from spyder_kernels.comms.frontendcomm import FrontendComm
+from spyder_kernels.py3compat import to_text_string
```
ccordoba12: This import is not necessary anymore.

PR #207, opened 2020-02-18
PR: Move CI to github actions on branch master

goanpeca:


goanpeca:
Need to think how to collect all coverage into a single batch, but that is work for another PR

ccordoba12:
> Need to think how to collect all coverage into a single batch, but that is work for another PR

Right, no problem about it. Let's merge this one.

.github/workflows/linux-tests.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     strategy:
       fail-fast: false 
       matrix:
-        PYTHON_VERSION: ['2.7', '3.6', '3.7', '3.8']
```
ccordoba12: I'd like to maintain compatibility with Python 2 for a little longer here so people can still evaluate their Python 2 code. Since that doesn't represent a big effort for this small package, please revert these changes.

goanpeca: Done

setup.py
```diff
@@ -38,7 +38,6 @@ def get_version(module='spyder_kernels'):
 REQUIREMENTS = [
     'cloudpickle',
     'ipykernel',
-    'ipykernel>=5.1.3; python_version>"2"',
```
goanpeca: @ccordoba12 what should I leave here?

ccordoba12: The same, and please also revert the classifier deletions you did below.

goanpeca: Ok

PR #208, opened 2020-02-22
PR: Add flag to override data and handle NpzFile instances

dalthviz:
Part of spyder-ide/spyder#11596


spyder_kernels/console/kernel.py
```diff
@@ -252,10 +252,11 @@ def load_data(self, filename, ext):
         if error_message:
             return error_message
 
-        for key in list(data.keys()):
-            new_key = fix_reference_name(key, blacklist=list(glbs.keys()))
-            if new_key != key:
-                data[new_key] = data.pop(key)
+        if not override:
```
goanpeca: Could you please add a comment explaining what this is actually doing now?

It is not so clear to me.

spyder_kernels/console/kernel.py
```diff
@@ -239,7 +239,7 @@ def copy_value(self, orig_name, new_name):
         ns = self._get_reference_namespace(orig_name)
         ns[new_name] = ns[orig_name]
 
-    def load_data(self, filename, ext):
+    def load_data(self, filename, ext, override=False):
         """Load data from filename"""
```
goanpeca: Please add a period at the end of the string. Also I think it is a good opportunity to explain what override  does and does not do.

spyder_kernels/console/kernel.py
```diff
@@ -252,10 +252,11 @@ def load_data(self, filename, ext):
         if error_message:
             return error_message
 
-        for key in list(data.keys()):
-            new_key = fix_reference_name(key, blacklist=list(glbs.keys()))
-            if new_key != key:
-                data[new_key] = data.pop(key)
+        if not override:
+            for key in list(data.keys()):
```
goanpeca: Why do  we convert to list?

`for key in data.keys():`this should work just fine, right?

spyder_kernels/console/kernel.py
```diff
@@ -252,10 +252,11 @@ def load_data(self, filename, ext):
...
+                new_key = fix_reference_name(key, blacklist=list(glbs.keys()))
```
goanpeca: Here we do need to keep the list() since keys() on newer python versions does not return a list.

spyder_kernels/utils/iofuncs.py
```diff
@@ -199,7 +199,9 @@ def load_array(filename):
         try:
             name = osp.splitext(osp.basename(filename))[0]
             data = np.load(filename)
-            if hasattr(data, 'keys'):
+            if isinstance(data, np.lib.npyio.NpzFile):
+                return dict(data), None
```
goanpeca: Please add a test for this specific case :-)

spyder_kernels/console/kernel.py
goanpeca: @dalthviz ?

dalthviz:
Note: Seems like the AppVeyor tests for Python 2.7 and 3.6 are running with an outdated version of IPython:

* Python 2.7: IPython `5.3.0` instead of `5.8.0` is being used.
* Python 3.6: IPython `6.1.0` instead of `7.12.0` is being used.

spyder_kernels/console/kernel.py
goanpeca: so this does not work

`for key in data.keys():`  ?

dalthviz: I was just checking if this change somehow affected the tests in AppVeyor but is unrelated

dalthviz:
Seems like the issue was caused because the testing env somehow started using pip `9.0.1` instead of a more recent pip version like `19.x` or `20.x`

dalthviz:
@goanpeca @ccordoba12 seems like the checks passed but I'm unsure if there is a better approach to handle the `pip` strange version change

requirements/windows.txt
```diff
@@ -1,2 +1,3 @@
 ipykernel
-cloudpickle
\ No newline at end of file
+cloudpickle
+pip>=19.3.1
```
goanpeca: I think this approach is fine for now, let's add a comment here to explain why we need the pin. Then we are good to go!

goanpeca:
@ccordoba12 this one is ready!

PR #220, opened 2020-05-04
PR: Make multithreading patch work for all OSes in Python 3

steff456:
Add the multithreading patch for fixing Dask issues if working in Python 3.
Also add a regression test

Fixes https://github.com/spyder-ide/spyder/issues/12465

### DIFF
--- a/requirements/tests.txt
+++ b/requirements/tests.txt
@@ -1,5 +1,6 @@
 codecov
 cython
+dask
 flaky
 matplotlib
 mock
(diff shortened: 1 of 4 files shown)


requirements/tests.txt
```diff
@@ -1,5 +1,6 @@
 codecov
 cython
+dask
```
ccordoba12, line 3: Please also add this new dependency to our `setup.py`.

spyder_kernels/console/tests/test_console_kernel.py
```diff
@@ -435,6 +435,49 @@ def f(x):
...
+from dask.distributed import Client
+
+if __name__=='__main__':
+    client = Client()
+    client.close()
+    x = 'hello'
+
+    client = Client()
+    client.close()
+    x = 'hello'
```
ccordoba12, line 465: Please remove these lines and instead run the file twice with `runfile` to exactly reproduce the conditions specified by the user on the Spyder issue.

PR #224, opened 2020-05-26
PR: Enable use of actual value of basic numeric types (Variable Explorer)

dalthviz:
Part of spyder-ide/spyder#12620

### DIFF
--- a/spyder_kernels/utils/nsview.py
+++ b/spyder_kernels/utils/nsview.py
@@ -292,7 +292,7 @@ def default_display(value, with_module=True):
         return type_str[1:-1]
 
 
-def collections_display(value, level, numeric_value=False):
+def collections_display(value, level):
     """Display for collections (i.e. list, set, tuple and dict)."""
     is_dict = isinstance(value, dict)
     is_set = isinstance(value, set)
@@ -315,15 +315,11 @@ def collections_display(value, level, numeric_value=False):
     # Get display of each element
     if level <= 2:
         if is_dict:
-            displays = [value_to_display(
-                            k, level=level, numeric_value=numeric_value)
-                        + ':' +
-                        value_to_display(
-                            v, level=level, numeric_value=numeric_value)
+            displays = [value_to_display(k, level=level) + ':' +
+                        value_to_display(v, level=level)
                         for (k, v) in list(elements)]
         else:
-            displays = [value_to_display(
-                            e, level=level, numeric_value=numeric_value)
+            displays = [value_to_display(e, level=level)
                         for e in elements]
         if truncate:
             displays.append('...')
@@ -449,7 +445,9 @@ def value_to_display(value, minmax=False, level=0, numeric_value=False):
         elif (isinstance(value, datetime.date) or
               isinstance(value, datetime.timedelta)):
             display = str(value)
-        elif isinstance(value, NUMERIC_TYPES) and numeric_value:
+        elif ((isinstance(value, NUMERIC_TYPES) or
+               isinstance(value, numeric_numpy_types)) and
+              numeric_value):
             display = value
         elif (isinstance(value, NUMERIC_TYPES) or
               isinstance(value, bool) or
--- a/spyder_kernels/utils/nsview.py
+++ b/spyder_kernels/utils/nsview.py
@@ -463,7 +463,9 @@ def value_to_display(value, minmax=False, level=0, numeric_value=False):
 
     # Truncate display at 70 chars to avoid freezing Spyder
     # because of large displays
-    if not isinstance(display, NUMERIC_TYPES) and len(display) > 70:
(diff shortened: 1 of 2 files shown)


spyder_kernels/utils/nsview.py
```diff
@@ -292,7 +292,7 @@ def default_display(value, with_module=True):
         return type_str[1:-1]
 
 
-def collections_display(value, level):
+def collections_display(value, level, numeric_value=False):
```
ccordoba12, line 295: I don't think it's necessary to add `numeric_value` as a kwarg to this function because its purpose is to return a string representation of a collection.

spyder_kernels/utils/nsview.py
```diff
@@ -445,6 +449,8 @@ def value_to_display(value, minmax=False, level=0):
         elif (isinstance(value, datetime.date) or
               isinstance(value, datetime.timedelta)):
             display = str(value)
+        elif isinstance(value, NUMERIC_TYPES) and numeric_value:
+            display = value
         elif (isinstance(value, NUMERIC_TYPES) or
               isinstance(value, bool) or
               isinstance(value, numeric_numpy_types)):
```
ccordoba12, line 456: I think adding `numeric_value` is not necessary if you simply rewrite these lines like this:

```python
elif (isinstance(value, NUMERIC_TYPES) or 
      isinstance(value, numeric_numpy_types)):
    display = value
elif isinstance(value, bool):
    display = repr(value)
```

I included `numeric_numpy_types` so people can sort them correctly too.

spyder_kernels/utils/nsview.py
```diff
@@ -459,7 +465,7 @@ def value_to_display(value, minmax=False, level=0):
 
     # Truncate display at 70 chars to avoid freezing Spyder
     # because of large displays
-    if len(display) > 70:
+    if not isinstance(display, NUMERIC_TYPES) and len(display) > 70:
```
ccordoba12, line 468: If you agree with my suggestion above, this needs to include `numeric_numpy_types` too.

spyder_kernels/utils/nsview.py
ccordoba12: There's an important thing I just verified about this: `complex` is in `NUMERIC_TYPES`, but complex numbers can't be sorted against other numbers (be them complex or not):

```
(2 + 3j) > 0
TypeError: '>' not supported between instances of 'complex' and 'int'

(2 + 3j) > (3 + 2j) 
TypeError: '>' not supported between instances of 'complex' and 'complex'
```

However, Numpy complex types can be sorted correctly. I also verified that there are no problems with `np.bool_`. So, I think we need to remove `complex` numbers at this point.

dalthviz: > 
> 
> I think adding `numeric_value` is not necessary if you simply rewrite these lines like this:
> 
> ```python
> elif (isinstance(value, NUMERIC_TYPES) or 
>       isinstance(value, numeric_numpy_types)):
>     display = value
> elif isinstance(value, bool):
>     display = repr(value)
> ```
> 
> I included `numeric_numpy_types` so people can sort them correctly too.

The reason to add the `numeric_value` param is that without it the value shown for the list in the Variable Explorer changes to `list object` instead of the current representation that is `[1e+16, 10.0, 1.0, 0.1, 1e-06, 0.0, -1e-06, -1.0, -10.0, -1e+16]` (`collections_display` uses `value_to_display` to get the representation of each element of the collection/list so that was causing the representation to change when I tested a simpler approach as the suggested)

dalthviz: > 
> 
> There's an important thing I just verified about this: `complex` is in `NUMERIC_TYPES`, but complex numbers can't be sorted against other numbers (be them complex or not):
> 
> ```
> (2 + 3j) > 0
> TypeError: '>' not supported between instances of 'complex' and 'int'
> 
> (2 + 3j) > (3 + 2j) 
> TypeError: '>' not supported between instances of 'complex' and 'complex'
> ```
> 
> However, Numpy complex types can be sorted correctly. I also verified that there are no problems with `np.bool_`. So, I think we need to remove `complex` numbers at this point.

Should I remove `complex` from the `NUMERIC_TYPES` constant then?

ccordoba12: > The reason to add the numeric_value param is that without it the value shown for the list in the Variable Explorer changes to list object instead of the current representation that is [1e+16, 10.0, 1.0, 0.1, 1e-06, 0.0, -1e-06, -1.0, -10.0, -1e+16] (collections_display uses value_to_display to get the representation of each element of the collection/list so that was causing the representation to change when I tested a simpler approach as the suggested)

Ok, understood.

> Should I remove complex from the NUMERIC_TYPES constant then?

Yeah, I think that's fine.

PR #230, opened 2020-07-02
PR: Send comm config before any wait just to be sure

impact27:
Send comm config any time we wait

### DIFF
--- a/.github/workflows/windows-tests.yml
+++ b/.github/workflows/windows-tests.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       fail-fast: false 
       matrix:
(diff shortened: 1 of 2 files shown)


impact27:
The windows 2.7 test is broken

ccordoba12:
Yeah, but it's an installation issue not a failure in our tests. So please remove `2.7` from here

https://github.com/spyder-ide/spyder-kernels/blob/0b77ca5b535c409793365643436b8290a28e1db0/.github/workflows/windows-tests.yml#L22

so we can merge this PR.

I really don't have time to see how to fix that and Python 2 is going to be removed in Spyder 4.2, so there's no need to waste on it.

PR #232, opened 2020-07-04
PR: Make do_where go to the current file

impact27:
Modify the behaviour or the `where` pdb command to go to the current place in the editor

### DIFF
--- a/spyder_kernels/customize/spyderpdb.py
+++ b/spyder_kernels/customize/spyderpdb.py
@@ -159,6 +159,18 @@ class SpyderPdb(ipyPdb, object):  # Inherits `object` to call super() in PY2
             return False
         return True
 
+    def do_where(self, arg):
+        """w(here)
+        Print a stack trace, with the most recent frame at the bottom.
+        An arrow indicates the "current frame", which determines the
+        context of most commands. 'bt' is an alias for this command.
+
+        Take a number as argument as an (optional) number of context line to
+        print"""
+        super(SpyderPdb, self).do_where(arg)
+        frontend_request().do_where()
+
+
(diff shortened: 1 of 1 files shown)


spyder_kernels/customize/spyderpdb.py
```diff
@@ -139,8 +139,6 @@ def interaction(self, frame, traceback):
             self.onecmd('exit')
         else:
             self.setup(frame, traceback)
-            if self.send_initial_notification:
-                self.notify_spyder(frame)
```
ccordoba12, line 143: Why is it necessary to move this to `preloop`?

spyder_kernels/customize/spyderpdb.py
```diff
@@ -322,8 +337,7 @@ def postcmd(self, stop, line):
 
         Is that good or too lazy? i.e. is more specific behaviour desired?
         """
-        if '!get_ipython().kernel' not in line:
```
ccordoba12, line 325: Why is this not necessary anymore?

impact27: This was the old way of running cells and is not used anymore

spyder_kernels/customize/spyderpdb.py
impact27: Otherwise the first prompt is not correctly handled by the where command

impact27: The first prompt is not the result of an interaction, so the initial notification is not sent unless I move this to `preloop`. Otherwise the initial interaction is only sent for the second prompt.

ccordoba12 (approved):
Thanks for the clarifications @impact27.

PR #238, opened 2020-07-15
PR: Set get_local_scope correctly while debugging

impact27:
While debugging, the `%timeit` magic doesn't set the local scope correctly. Together with https://github.com/ipython/ipython/pull/12437, this PR solves this problem by correctly setting the locals while debugging.


ccordoba12:
Does this generate if IPython version is less than 7.17? Or is there no problem with that?

impact27:
> Does this generate an error if the IPython version is less than 7.17? Or is there no problem with that?

Get_local_scope doesn’t do anything before 7.17 so it should be fine as this is just adding an unused method

PR #240, opened 2020-07-18
PR: Hide runcell and runfile frames using a new IPython feature

impact27:
Use the new feature introduced in https://github.com/ipython/ipython/pull/12359 to hide spyder internals

### DIFF
--- a/spyder_kernels/customize/spydercustomize.py
+++ b/spyder_kernels/customize/spydercustomize.py
@@ -400,6 +400,8 @@ def transform_cell(code):
 
 def exec_code(code, filename, ns_globals, ns_locals=None, post_mortem=False):
     """Execute code and display any exception."""
+    # Tell IPython to hide this frame
+    __tracebackhide__ = True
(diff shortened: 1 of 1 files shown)


spyder_kernels/customize/spydercustomize.py
```diff
@@ -400,6 +400,8 @@ def transform_cell(code):
 
 def exec_code(code, filename, ns_globals, ns_locals=None, post_mortem=False):
     """Execute code and display any exception."""
+    # Tell IPython to hide this frame
```
dalthviz, line 403: Could be worthy to add here (and the other places below where the hiding is set) a reference to the IPython version that starts supporting this?

dalthviz (commented):
Thanks @impact27 ! I left a simple comment. Also I was wondering if this could be tested somehow and maybe a rebase could be worthy? What do you think? Besides that, this LGTM 👍

dalthviz (approved):
Thanks @impact27 after checking a little bit more ipython/ipython#12359 this LGTM :+1:

PR #243, opened 2020-09-05
PR: Allow leading indents in cells

### LINKED ISSUE #247: Remove from warning that future versions of Spyder will give an error when using non-Python syntax
Currently we're showing this warning

immagine

I think we should remove its last sentence and perhaps reword its last part. That's because it seems a lot of people depends on this feature and forcing them to use `ipy` files would be a nuisance.

@impact27, what do you think?

impact27:
Fixes #247

### DIFF
--- a/spyder_kernels/customize/spydercustomize.py
+++ b/spyder_kernels/customize/spydercustomize.py
@@ -20,6 +20,7 @@ import sys
 import time
 import warnings
 import logging
+import re
 
 from IPython.core.getipython import get_ipython
 
@@ -30,7 +31,8 @@ from spyder_kernels.customize.spyderpdb import SpyderPdb
 from spyder_kernels.customize.umr import UserModuleReloader
 
 if not PY2:
-    from IPython.core.inputtransformer2 import TransformerManager
+    from IPython.core.inputtransformer2 import (
+        TransformerManager, leading_indent, leading_empty_lines)
 else:
     from IPython.core.inputsplitter import IPythonInputSplitter as TransformerManager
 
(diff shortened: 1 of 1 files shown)


spyder_kernels/customize/spydercustomize.py
```diff
@@ -30,7 +31,8 @@
 from spyder_kernels.customize.umr import UserModuleReloader
 
 if not PY2:
```
ccordoba12, line 33: The `inputtransformer2` module was introduced in IPython 7, so this condition needs to be changed in the same way as I did in https://github.com/spyder-ide/spyder/pull/13721

ccordoba12: Probably the `if`s you modified below need to be adjusted in the same way, but I'm not really sure.

ccordoba12 (commented):
@impact27, I left a minor comment for you, otherwise looks good to me.

Could you also add a test for this? Thanks!

impact27:
Fixes https://github.com/spyder-ide/spyder-kernels/issues/247

spyder_kernels/customize/spydercustomize.py
impact27: should be fixed

impact27:
Test implemented in https://github.com/spyder-ide/spyder/pull/13852

spyder_kernels/customize/spydercustomize.py
```diff
@@ -442,9 +458,8 @@ def exec_code(code, filename, ns_globals, ns_locals=None, post_mortem=False):
                             "\nWARNING: This is not valid Python code. "
                             "If you want to use IPython magics, "
                             "flexible indentation, and prompt removal, "
-                            "please save this file with the .ipy extension. "
-                            "This will be an error in a future version of "
-                            "Spyder.\n")
+                            "we recommand that you save this file with the "
```
ccordoba12, line 461: recommand -> recommend

ccordoba12 (commented):
I left a small suggestion, otherwise looks good.

spyder_kernels/customize/spydercustomize.py
```diff
@@ -22,6 +22,8 @@
 import logging
 import cmd
 
+from distutils.version import LooseVersion
```
ccordoba12, line 25: `distutils` is part of the standard library, so it needs to go in the previous section of imports. Could you also organize them in alphabetical order as part of this PR? Thanks!

ccordoba12 (commented):
Sorry @impact27, i missed another small thing.

ccordoba12 (approved):
Thanks @impact27!

PR #244, opened 2020-09-15
PR: Set multiprocessing ORIGINAL_DIR at startup

impact27:
sets "Original wdir" before running a file

Fixes spyder-ide/spyder#13632

### DIFF
--- a/spyder_kernels/customize/spydercustomize.py
+++ b/spyder_kernels/customize/spydercustomize.py
@@ -522,14 +522,20 @@ def runfile(filename=None, args=None, wdir=None, namespace=None,
             for arg in shlex.split(args):
                 sys.argv.append(arg)
         if wdir is not None:
-            try:
-                wdir = wdir.decode('utf-8')
-            except (UnicodeError, TypeError, AttributeError):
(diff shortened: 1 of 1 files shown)


spyder_kernels/customize/spydercustomize.py
```diff
@@ -522,14 +522,22 @@ def runfile(filename=None, args=None, wdir=None, namespace=None,
...
-            try:
-                wdir = wdir.decode('utf-8')
-            except (UnicodeError, TypeError, AttributeError):
-                # UnicodeError, TypeError --> eventually raised in Python 2
-                # AttributeError --> systematically raised in Python 3
-                pass
+            if PY2:
+                try:
+                    wdir = wdir.decode('utf-8')
+                except (UnicodeError, TypeError, AttributeError):
```
dalthviz, line 528: Since this block now is only run with Python 2 should we still keep the `AttributeError` in the except?

dalthviz (commented):
Thanks @impact27 ! Left a comment but besides that this LGTM +1:

dalthviz (approved):
Thanks @impact27 LGTM :+1:

PR #246, opened 2020-09-16
PR: Allow sorting with custom sorting key

skjerns:
related to https://github.com/skjerns/spyder/tree/issue_13733

should not alter any current functionality and be backwards compatible.


dalthviz (approved):
Thanks @skjerns LGTM :+1:

PR #249, opened 2020-09-20
PR: Show size of objects correctly, if they implement len()

### LINKED ISSUE #245: Display size of string in variable editor
Currently the variable explorer displays all strings with size `1` - I think it would be more intuitive to show the length of the string.

It would simply requiring to change the size here to include `str`

https://github.com/spyder-ide/spyder-kernels/blob/87d89ad94becb51f4eea1bea643334fd02440d6c/spyder_kernels/utils/nsview.py#L124

skjerns:
#245  I suggested to show the string size correctly and not display `1`.

I thought I even extend that to all objects implementing `len()`. I think if the object can tell us how large it is, we can display this self-reported size.

Fixes #245.

### DIFF
--- a/spyder_kernels/utils/tests/test_nsview.py
+++ b/spyder_kernels/utils/tests/test_nsview.py
@@ -73,7 +73,7 @@ def test_get_size():
     assert get_size(img) == (256,256)
 
     obj = RecursionClassNoLen()
-    assert get_size(obj) == -1
+    assert get_size(obj) == 1
 
 
 
--- a/spyder_kernels/utils/tests/test_nsview.py
+++ b/spyder_kernels/utils/tests/test_nsview.py
@@ -18,12 +18,13 @@ import numpy as np
 import pandas as pd
 import pytest
 import xarray as xr
+import PIL
 
 # Local imports
 from spyder_kernels.py3compat import PY2
 from spyder_kernels.utils.nsview import (sort_against, is_supported,
-                                         value_to_display,
-                                         get_supported_types)
+                                         value_to_display, get_size,
+                                         get_supported_types, Image)
 
 def generate_complex_object():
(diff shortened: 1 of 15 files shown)


dalthviz (commented):
Thanks @skjerns ! I think we could add a test for this, could you please add one? (you can add it [here](https://github.com/spyder-ide/spyder-kernels/blob/1.x/spyder_kernels/utils/tests/test_nsview.py)), otherwise this LGTM :+1:

skjerns:
I've added tests. Additionally, I extended the functionality such that it will show either `x.shape`, `x.size` or `len(x)`. I think this should cover 98% of objects that have a length.

spyder_kernels/utils/tests/test_nsview.py
```diff
@@ -18,12 +18,13 @@
 import pandas as pd
 import pytest
 import xarray as xr
+import PIL
```
ccordoba12, line 21: Please add `pillow` to our requirements for our tests in our `setup.py` and `requirements/tests.txt`.

ccordoba12 (commented):
Hey @skjerns, thanks for the new test. Unfortunately you merged `master` with this branch, which was incorrect because this PR needs to be against our `1.x` branch.

Could you fix that?

skjerns:
> Hey @skjerns, thanks for the new test. Unfortunately you merged `master` with this branch, which was incorrect because this PR needs to be against our `1.x` branch.
> 
> Could you fix that?

Oh no, not again 😅 I'll try to fix it...

ccordoba12:
> It might be preferable to make a new branch and open a new PR for this. What do you think?

Totally agree, if that's not much to ask.

skjerns:
no problem! much easier than rebasing :) continued at #250

PR #214, opened 2020-04-04
PR: Use '!' for Pdb commands

impact27:
Use '!' before Pdb commands and nothing before python code, as is the case in IPython

PR #251, opened 2020-09-30
PR: Catch errors when trying to get the len of an object

impact27:
This address https://github.com/spyder-ide/spyder/issues/13891 and spyder-ide/spyder#13892

having __len__ doesn't mean len will return without errors

### DIFF
--- a/spyder_kernels/utils/nsview.py
+++ b/spyder_kernels/utils/nsview.py
@@ -118,28 +118,33 @@ def try_to_eval(value):
     except (NameError, SyntaxError, ImportError):
         return value
 
-    
+
 def get_size(item):
     """Return shape/size/len of an item of arbitrary type"""
-    if hasattr(item, 'shape') and isinstance(item.shape, (tuple, integer)):
-        try:
-            return item.shape
-        except RecursionError:
-            # This is necessary to avoid an error when trying to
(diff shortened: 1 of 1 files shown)


impact27:
Folded into https://github.com/spyder-ide/spyder-kernels/pull/214 to fix the tests

impact27:
Fixes https://github.com/spyder-ide/spyder/issues/13892 and the spyder `spyder/plugins/variableexplorer/widgets/tests/test_collectioneditor.py::test_sort_numpy_numeric_collectionsmodel` test

dalthviz (commented):
Thanks @impact27 ! maybe expanding the test to add the case where `len` raises an error could be worthy?
https://github.com/spyder-ide/spyder-kernels/blob/cb11f77cee7c79dad2cd7ed3c0e1aee1aa628164/spyder_kernels/utils/tests/test_nsview.py#L45

Besides that this LGTM :+1:

ccordoba12:
> maybe expanding the test to add the case where len raises an error could be worthy?

Thanks for the review @dalthviz! Given that @impact27 added an extra try/except that will catch any error generated by the `get_size` function, I think that's not worthy.

ccordoba12 (approved):
Thanks @impact27!

PR #252, opened 2020-10-03
PR: Use frame locals when evaluating code in Pdb

impact27:
Apply fix in https://github.com/python/cpython/pull/15194 to spyder-kernels

This addreses https://github.com/spyder-ide/spyder/issues/13909

### DIFF
--- a/spyder_kernels/customize/spyderpdb.py
+++ b/spyder_kernels/customize/spyderpdb.py
@@ -74,7 +74,8 @@ class SpyderPdb(ipyPdb, object):  # Inherits `object` to call super() in PY2
                        " command instead")
             return
         locals = self.curframe_locals
-        globals = self.curframe.f_globals
+        ns = self.curframe.f_globals.copy()
+        ns.update(locals)
         try:
             line = TransformerManager().transform_cell(line)
             try:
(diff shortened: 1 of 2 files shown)


spyder_kernels/customize/spyderpdb.py
```diff
@@ -74,7 +74,8 @@ def default(self, line):
                        " command instead")
             return
         locals = self.curframe_locals
-        globals = self.curframe.f_globals
+        ns = self.curframe.f_globals.copy()
+        ns.update(locals)
```
ccordoba12, line 78:
```suggestion

        # This is necessary to allow running comprehensions with the
        # frame locals. It also fallbacks to the right globals if the
        # user wants to work with them instead.
        # See spyder-ide/spyder#13909.
        if not 'globals()' in line:
            ns = self.curframe.f_globals.copy()
            ns.update(locals)
        else:
            ns = self.curframe.f_globals

```

ccordoba12 (commented):
Thanks @impact27 for this! I think with my suggestion below we'll fallback to the right globals in case the user wants to work with them instead.

(Well, it won't cover exotic cases of writing `globals()`, but I think it should be enough for 99% of the use cases).

ccordoba12 (approved):
Thanks @impact27! I also added a test for this, so we don't accidentally break this feature in the future.

PR #255, opened 2020-11-01
PR: Fix warning for '!' Pdb option

impact27:
The warning was being printed when not using pdb commands


ccordoba12 (approved):
Thanks @impact27!

PR #262, opened 2020-11-25
PR: Avoid infinite hang when quitting external kernels

impact27:
Close the comms thread to avoid hanging while quitting. To reproduce:
`python -m spyder-kernels.console`
Connect to the console, send `quit()`
See that the console doesn't quit


ccordoba12 (approved):
Thanks @impact27, looks good to me.

PR #266, opened 2020-12-02
PR: Add exception for quit and exit in SpyderPdb

impact27:
If a user want to quit the debugging, he might write quit or exit. Avoid that doing that closes the kernel instead.


ccordoba12 (approved):
Thanks @impact27, looks good to me.

PR #271, opened 2020-12-22
PR: Fix recursive debugger

impact27:
Addresses https://github.com/spyder-ide/spyder/issues/14472

### DIFF
--- a/spyder_kernels/customize/spydercustomize.py
+++ b/spyder_kernels/customize/spydercustomize.py
@@ -27,7 +27,7 @@ from IPython.core.getipython import get_ipython
 
 from spyder_kernels.comms.frontendcomm import CommError, frontend_request
 from spyder_kernels.customize.namespace_manager import NamespaceManager
-from spyder_kernels.customize.spyderpdb import SpyderPdb
+from spyder_kernels.customize.spyderpdb import SpyderPdb, enter_debugger
 from spyder_kernels.customize.umr import UserModuleReloader
 from spyder_kernels.py3compat import TimeoutError, PY2, _print, encode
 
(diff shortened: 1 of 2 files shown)


spyder_kernels/customize/spydercustomize.py
```diff
@@ -583,6 +571,54 @@ def runfile(filename=None, args=None, wdir=None, namespace=None,
 builtins.runfile = runfile
 
 
+def enter_debugger(filename, continue_if_has_breakpoints, code_format):
```
ccordoba12, line 574: Given that this function is quite long now, I'd prefer to move it to `spyder_kernels/customize/spyderpdb.py`

spyder_kernels/customize/spydercustomize.py
```diff
@@ -583,6 +571,54 @@ def runfile(filename=None, args=None, wdir=None, namespace=None,
...
+    debugger._user_requested_quit = False
+
+    if os.name == 'nt':
+        filename = filename.replace('\\', '/')
+
+    code = code_format.format(repr(filename))
+
+    if recursive:
+        # Enter recursive debugger
+        parent_debugger.message("ENTERING RECURSIVE DEBUGGER")
```
ccordoba12, line 610: Are you sure you want to use all caps for this message? I mean, is it really such an important message for that?

spyder_kernels/customize/spydercustomize.py
```diff
@@ -583,6 +571,54 @@ def runfile(filename=None, args=None, wdir=None, namespace=None,
...
+        sys.call_tracing(debugger.run, (code, globals, locals))
+        parent_debugger.message("LEAVING RECURSIVE DEBUGGER")
```
ccordoba12, line 612: The same applies here, of course.

ccordoba12 (commented):
Thanks @impact27 for your help with this! I left some small comments for you.

spyder_kernels/customize/spydercustomize.py
impact27: This is to be consistant with https://github.com/ipython/ipython/blob/106a841993ee773fbe4045b8aec5cb37d6cc3eaa/IPython/core/debugger.py#L606, But I think we can drop the message.

ccordoba12 (approved):

Issue #274, opened 2021-01-01
Five failing tests


juliangilbey:
I've been trying to build the current spyder-kernels (1.10.1) and three of the tests are failing for different reasons.

The first is `test_turtle_launch`, which is protected by a check for tkinter being installed, but this doesn't actually check whether windows can be drawn.  It's not much of an issue, but when run in a non-interactive environment where tkinter happens to be installed, this fails.  It might be possible to do a slightly better test than importing tkinter, but I don't know what that would be: perhaps trying to draw a window or something?

The second is `test_dask_multiprocessing` which fails for a reason I can't fathom, perhaps due to the presence of certain python packages? But it's not a consistent failure.  Here is the log:

```python
__________________________ test_dask_multiprocessing ___________________________

tmpdir = local('/tmp/pytest-of-jdg/pytest-47/test_dask_multiprocessing2')

    @flaky(max_runs=3)
    @pytest.mark.skipif(not PY3,
                        reason="Only meant for Python 3")
    def test_dask_multiprocessing(tmpdir):
        """
        Test that dask multiprocessing works on Python 3.
        """
        # Command to start the kernel
        cmd = "from spyder_kernels.console import start; start.main()"
    
        with setup_kernel(cmd) as client:
            # Remove all variables
            client.execute("%reset -f")
            client.get_shell_msg(block=True, timeout=TIMEOUT)
    
            # Write multiprocessing code to a file
            # Runs two times to verify that in the second case it doesn't break
            code = """
    from dask.distributed import Client
    
    if __name__=='__main__':
        client = Client()
        client.close()
        x = 'hello'
    """
            p = tmpdir.join("mp-test.py")
            p.write(code)
    
            # Run code two times
            client.execute("runfile(r'{}')".format(to_text_string(p)))
>           client.get_shell_msg(block=True, timeout=TIMEOUT)

tests/console/tests/test_console_kernel.py:468: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/lib/python3/dist-packages/jupyter_client/client.py:78: in get_shell_msg
    return self.shell_channel.get_msg(*args, **kwargs)
[...]

PR #275, opened 2021-01-03
PR: Better test for non-interactive context

juliangilbey:
This addresses the "easy" part of #274 by testing for the ability to run turtle explicitly (in case tkinter is installed but the test is running in a non-graphical context).

### DIFF
--- a/spyder_kernels/console/tests/test_console_kernel.py
+++ b/spyder_kernels/console/tests/test_console_kernel.py
@@ -43,13 +43,6 @@ FILES_PATH = os.path.dirname(os.path.realpath(__file__))
 TIMEOUT = 15
 SETUP_TIMEOUT = 60
 
-TKINTER_INSTALLED = False
-try:
-    import tkinter
-    TKINTER_INSTALLED = True
-except:
-    pass
-
 
 @contextmanager
 def setup_kernel(cmd):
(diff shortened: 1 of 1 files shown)


ccordoba12:
Hey @juliangilbey, please move this PR to be against our 1.x (using the same commands I mentioned to you in Spyder) and rebase it on top of fastet so you can get the fix to our tests.

juliangilbey:
Done!

spyder_kernels/console/tests/test_console_kernel.py
```diff
@@ -605,9 +598,18 @@ def test_np_threshold(kernel):
...
 
 
+TURTLE_ACTIVE = False
+try:
+    import turtle
+    turtle.Screen()
+    turtle.bye()
+    TURTLE_ACTIVE = True
+except:
+    pass
```
ccordoba12, line 608: Please move this block to be where the `TKINTER_INSTALLED` one was present. We prefer to declare constants on top and not in the middle of tests.

ccordoba12 (commented):
One small comment, then this should be ready @juliangilbey.

spyder_kernels/console/tests/test_console_kernel.py
juliangilbey: I put it there, because if running interactively, it pops up a window to test for this, and I thought it made more sense to do this immediately before the test it relates to. But I'm happy to move it if you prefer.

ccordoba12: Ok, I see. Then please add a comment on top of this block explaining that.

juliangilbey: Done!

ccordoba12 (approved):
Looks good to me now, thanks @juliangilbey!

PR #278, opened 2021-01-11
PR: Warn when comm call creates text output

impact27:
Addresses https://github.com/spyder-ide/spyder/issues/14551

```
In [1]: class A():
   ...:     @property
   ...:     def shape(self):
   ...:         print((1234, 1))
   ...:         return 1234, 1

In [2]: a = A()

Output from spyder call 'get_namespace_view':
(1234, 1)
(1234, 1)
(1234, 1)
(1234, 1)

In [3]: 
```

### DIFF
--- a/spyder_kernels/comms/frontendcomm.py
+++ b/spyder_kernels/comms/frontendcomm.py
@@ -13,7 +13,6 @@ import socket
 import sys
 import threading
 import time
-from io import TextIOBase
 
 from jupyter_client.localinterfaces import localhost
 from tornado import ioloop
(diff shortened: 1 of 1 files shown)


ccordoba12 (commented):
Hey @impact27, sorry for the late review. Could you add a test to check that this fixes the issue reported in the Spyder repo? Thanks!

impact27:
> 
> 
> Hey @impact27, sorry for the late review. Could you add a test to check that this fixes the issue reported in the Spyder repo? Thanks!

Done here:  https://github.com/spyder-ide/spyder/pull/14715

ccordoba12 (approved):
Thanks @impact27 for your help with this one!

PR #280, opened 2021-03-09
Added option to use magic in .py files

pinguicula:


ccordoba12:
Hey @pinguicula, thanks for your contribution. Could you explain us what you're trying to achieve with these changes?

impact27:
I think magic are only valid in .ipy files, because if you use magics in your file, then it is not a python file, but a ipython file. You can just rename your file (file.py -> file.ipy), unless you have a usecase where this is not possible?

pinguicula:
I am using spyder in my workflow for development of .py files. I use cells very frequently. Running cells are interpreted as flexible indentation which means spyder think that they are magic. Example:

```
for a in b:
#%% <- This cell is sometimes, but not always, not possible to run due to the flexible indentation.
    x = fun()
```

impact27:
I am not sure I follow:
```
b = [1, 2]
for a in b:
#%%
    print(a)
```
runs fine for me both in spyder and with python3.

`#%%` is a comment, not a magic, that spyder interprets as a cell break. This works both for .py and .ipy files.

A magic would be for example:
`%matplotlib`
Which wouldn't run in python.

pinguicula:
Cells are sometimes interpreted as flexible indentation which spyder thinks are magic.

impact27:
If you can reproduce I would be interested to see the code that triggers this behaviour.

Usually I would just use the console to write magics. I never had a problem running cells separated  by `#%%` even while debugging.

PR #287, opened 2021-04-29
PR: Close comm on shutdown

impact27:
If a shutdown message is recieved on the comm channel, close comm


ccordoba12 (approved):
Thanks @impact27!

PR #290, opened 2021-05-03
PR: Add constraints on IPython and ipykernel

ccordoba12:
- IPython 7.23 breaks the inline backend
- ipykernel 6 could break us because it comes with a lot of changes
- I removed the pin on decorator < 5 because the Cython magic doesn't fail with it anymore.


ccordoba12:
This is not longer necessary.

PR #298, opened 2021-06-01
PR: Use backslash instead of slash for UMR path regex detection

dalthviz:
Addresses https://github.com/spyder-ide/spyder/issues/15417


ccordoba12 (approved):

PR #308, opened 2021-08-17
fix globals

impact27:
Fixes https://github.com/spyder-ide/spyder/issues/16201

### DIFF
--- a/spyder_kernels/customize/namespace_manager.py
+++ b/spyder_kernels/customize/namespace_manager.py
@@ -55,6 +55,7 @@ class NamespaceManager(object):
         self._previous_main = None
         self._reset_main = False
         self._file_code = file_code
+        self._saved_globals = None
 
     def __enter__(self):
         """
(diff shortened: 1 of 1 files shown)


ccordoba12:
@impact27, I think the failures in our tests are related to this: https://github.com/ipython/ipykernel/pull/736

spyder_kernels/customize/namespace_manager.py
```diff
@@ -55,23 +55,36 @@ def __init__(self, filename, namespace=None, current_namespace=False,
...
                 if '__file__' in self.ns_globals:
                     self._previous_filename = self.ns_globals['__file__']
                 self.ns_globals['__file__'] = self.filename
             else:
                 ipython_shell = get_ipython()
+                shell_namespace = ipython_shell.user_ns
                 main_mod = ipython_shell.new_main_mod(
                     self.filename, '__main__')
-                self.ns_globals = main_mod.__dict__
+                if shell_namespace is global_namespace:
```
ccordoba12, line 78: How efficient is this comparison when there are thousands of variables in the namespace?

spyder_kernels/customize/namespace_manager.py
```diff
@@ -55,23 +55,36 @@ def __init__(self, filename, namespace=None, current_namespace=False,
...
+                    # Run in the global namespace
+                    self._saved_globals = shell_namespace.copy()
+                    # Here we clear all variables, let's hope ipykernel
+                    # is not too mad
+                    shell_namespace.clear()
```
ccordoba12, line 83: IPython saves several hidden variables that don't come from running code. This could break that.

spyder_kernels/customize/namespace_manager.py
impact27: I think the number of variables doesn’t impact the comparison speed, my understanding is that you are just checking if the pointers are the same.

spyder_kernels/customize/namespace_manager.py
impact27: Yes but there shouldn’t be any ipython code running between enter and exit. All the code runs in the same thread does it not? I agree that this is a risk

spyder_kernels/customize/namespace_manager.py
ccordoba12: Ok, good to know.

spyder_kernels/customize/namespace_manager.py
ccordoba12: > Yes but there shouldn’t be any ipython code running between enter and exit. All the code runs in the same thread does it not?

Yeah, I think so.

> I agree that this is a risk

Have you tested this for a while to see if it introduces problems?

impact27: I think we need to call `get_ipython().init_user_ns` after clearing the namespace to add the hidden variables back.

impact27:
I am not confident enough about not introducing a bug with that one.

PR #311, opened 2021-08-25
Remove deps on ipython_genutils

Carreau:
It requires nose in test suite, that is not package on many linux distrib
and should not be used anymore.

Technically this change is a bit more conservative than what genutils
was doing when trying to detect the default encoding, but that should
not matter much now.


ccordoba12:
Hey @Carreau, thanks a lot for your contribution! Please make your branch to derive from our `2.x` branch (instead of `master`), with the following commands:

    git checkout 2.x
    git checkout genu
    git rebase --onto 2.x master genu
    git push -f origin genu

Otherwise, looks good to me.

PR #312, opened 2021-08-25
PR: Use a yaml file to install conda dependencies on CIs

ccordoba12:
This allows us to have a single file to handle all our dependencies.


ccordoba12:
This doesn't work because env yaml files don't support selectors.

PR #328, opened 2021-10-25
PR: Remove upper constraint on jupyter_client

bnavigator:
According to the [migration guide](https://jupyter-client.readthedocs.io/en/master/migration.html), removing the `block` parameter should be all that is needed here. Let's see...


ccordoba12 (approved):
Thanks a lot @bnavigator for your help with this!

PR #329, opened 2021-11-01
PR: Fix debugger depth

impact27:
Fix debugger depth and cleanup the frontend notification
Fixes https://github.com/spyder-ide/spyder-kernels/issues/330

### DIFF
--- a/spyder_kernels/comms/frontendcomm.py
+++ b/spyder_kernels/comms/frontendcomm.py
@@ -35,7 +35,7 @@ def get_free_port():
     return port
 
 
-def frontend_request(blocking=True, timeout=None):
+def frontend_request(blocking, timeout=None):
     """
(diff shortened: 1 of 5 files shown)


ccordoba12:
Hey @impact27, how important do you think this is? I mean, is it worth to include it in our next release (2.2.0) that will be needed by Spyder 5.2.0?

impact27:
I would say this is a minor inconvenance but it makes the code nicer. It is needed for the profiler PR#286

ccordoba12:
@impact27, why did you decide to close this PR?

impact27:
I had to change the logic of exec_code to conditionally return something, and so I decided to fix #330 there. This one is not needed anymore when https://github.com/spyder-ide/spyder-kernels/pull/355 is merged

PR #332, opened 2021-11-09
PR: Remove usage of f_locals.get

impact27:
Fixes https://github.com/spyder-ide/spyder/issues/16780

### DIFF
--- a/spyder_kernels/customize/spyderpdb.py
+++ b/spyder_kernels/customize/spyderpdb.py
@@ -286,8 +286,8 @@ class SpyderPdb(ipyPdb, object):  # Inherits `object` to call super() in PY2
     def stop_here(self, frame):
         """Check if pdb should stop here."""
         if (frame is not None
-                and frame.f_locals.get(
-                    "__tracebackhide__", False) == "__pdb_exit__"):
+                and "__tracebackhide__" in frame.f_locals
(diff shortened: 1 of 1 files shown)


impact27:
See this:

That seems to be a bug in Pytorch.

According to [PEP-667](https://www.python.org/dev/peps/pep-0667/#python):

> `frame.f_locals` will return a view object on the frame that implements the `collections.abc.Mapping` interface.

And [`collections.abc.Mapping`](https://docs.python.org/3/library/collections.abc.html#:~:text=keys%2C%20items%2C%20values%2C-,get,-%2C%20__eq__%2C%20and%20__ne__) interface lists `get`:

> ABC | Inherits from | Abstract Methods | Mixin Methods
> -- | -- | -- | --
> Mapping | Collection | `__getitem__`, `__iter__`, `__len__` | `__contains__`, `keys`, `items`, `values`, `get`, `__eq__`, and `__ne__`

_Originally posted by @Kojoley in https://github.com/ipython/ipython/issues/13264#issuecomment-964743700_

ccordoba12:
As I mentioned in the corresponding PR @impact27 opened in IPython: that PEP is still in draft status. Also that PR was merged, so I think it's safe to merge this one too.

ccordoba12 (approved):
Thanks @impact27!

PR #333, opened 2021-11-09
PR: Add cast for shape tuple subclass instances when getting values size

dalthviz:
Addresses spyder-ide/spyder#16348


ccordoba12 (approved):

PR #334, opened 2021-11-10
Fix stderr missing

impact27:
stderr can be replaced before a crash, resulting in the error never being sent to the frontend. This explicitly writes the error to `__stderr__`. Worst case the error will be written twice.


ccordoba12 (approved):

PR #335, opened 2021-11-11
PR: Remove locals from exec with comprehensions (debugger)

impact27:
Addresses https://github.com/spyder-ide/spyder/issues/16790


ccordoba12 (commented):
Thanks for your help with this @impact27! It looks good to me, but could you add a test for this similar to `test_comprehensions_with_locals_in_pdb`?

https://github.com/spyder-ide/spyder-kernels/blob/2.x/spyder_kernels/console/tests/test_console_kernel.py#L803

ccordoba12 (approved):

PR #339, opened 2021-11-24
PR: Use control channel for comms

impact27:
Use control channels for comms now that it is in a thread


ccordoba12:
Thanks a lot for this @impact27! I think to simplify things, we should leave this change for spyder-kernels 3.0 because my plan is to drop support for Python 2 on that version.

spyder_kernels/comms/commbase.py
```diff
@@ -393,18 +393,20 @@ def on_incoming_call(self, call_dict):
...
-        """
+    def _send_call(self, call_dict, call_data, comm_id):
+        """Send call."""
         call_dict = self.on_outgoing_call(call_dict)
         self._send_message(
             'remote_call', content=call_dict, data=call_data,
             comm_id=comm_id)
 
+    def _get_call_return_value(self, call_dict, comm_id):
+        """_get_call_return_value
```
ccordoba12, line 404:
```suggestion
        """
```

No need to have the function name here.

spyder_kernels/comms/frontendcomm.py
```diff
@@ -128,25 +67,19 @@ def poll_one(self):
...
-        handler = self.kernel.shell_handlers.get(msg_type, None)
+        handler = self.kernel.control_handlers.get(msg_type, None)
         try:
             if handler is None:
                 self.kernel.log.warning("Unknown message type: %r", msg_type)
                 return
-            import asyncio
 
             if (getattr(asyncio, 'run', False) and
                     asyncio.iscoroutinefunction(handler)):
```
ccordoba12, line 85: You can remove this `if` statement and leave only `asyncio.run(handler(out_stream, ident, msg))` because we don't support Python 3.6 anymore.

spyder_kernels/comms/frontendcomm.py
```diff
@@ -213,18 +152,15 @@ def _comm_open(self, comm, msg):
...
-        call_dict["comm_port"] = self.comm_port
-        return super(FrontendComm, self).on_outgoing_call(call_dict)
-
-    def _send_comm_config(self):
-        """Send the comm config to the frontend."""
-        self.remote_call()._set_comm_port(self.comm_port)
+        self._set_pickle_protocol(
+            msg['content']['data']['pickle_highest_protocol'])
         self.remote_call()._set_pickle_protocol(pickle.HIGHEST_PROTOCOL)
+        # Handle cacahed messages
```
ccordoba12, line 158:
```suggestion
        # Handle cached messages
```

ccordoba12 (commented):
Thanks @impact27 for your work on this! I left a tiny review for you, otherwise looks good to me.

ccordoba12 (approved):

PR #352, opened 2021-12-22
PR: Put user provided paths at the front of sys.path

mrclary:
See spyder-ide/spyder#17066

### DIFF
--- a/spyder_kernels/customize/spydercustomize.py
+++ b/spyder_kernels/customize/spydercustomize.py
@@ -820,6 +820,12 @@ def set_spyder_pythonpath():
     pypath = os.environ.get('SPY_PYTHONPATH')
     if pypath:
         pathlist = pypath.split(os.pathsep)
+
+        if '' in sys.path:
+            # put cwd at front
+            sys.path.remove('')
+            pathlist.insert(0, '')
+
         # User provided paths should take priority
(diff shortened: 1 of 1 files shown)


spyder_kernels/customize/spydercustomize.py
```diff
@@ -727,6 +727,7 @@ def set_spyder_pythonpath():
     pypath = os.environ.get('SPY_PYTHONPATH')
     if pypath:
         pathlist = pypath.split(os.pathsep)
-        sys.path.extend(pathlist)
+        # User provided paths should take priority
+        [sys.path.insert(0, p) for p in reversed(pathlist)]
```
impact27, line 731: Why not `sys.path = pathlist + sys.path`?

impact27: Actually it probably doesn't matter, the number of paths is not that large

mrclary: `sys.path` should not be reassigned in that manner because `sys.path` is a mutable object that may have been imported somewhere else prior to being manipulated here. Any namespace that has already imported `sys.path` will continue to point to the original memory address location, while subsequent imports will point to the new `sys.path`, thus `sys.path` should be manipulated "in place". Both `.extend()` and `.insert()` methods do this.

PR #354, opened 2021-12-23
PR: Constraint IPython version to avoid issues when its next major version is released

ccordoba12:
We require now `ipython >=7.6.0,<8`

### DIFF
--- a/requirements/posix.txt
+++ b/requirements/posix.txt
@@ -1,3 +1,4 @@
 cloudpickle
 ipykernel
 wurlitzer>=1.0.3
+ipython<8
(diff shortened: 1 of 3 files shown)


eendebakpt:
@ccordoba12 IPython 8.,x has been out some time now and contains several bugfixes. Can we relax the version requirement for spyder to `ipython >=7.x.y`? If not, what still needs to be done?

ccordoba12:
We need to make sure that our tests both here and in Spyder pass with IPython 8 before doing that. And that will happen when we have time for it, which could take us a couple of months or more. It really depends on more pressing things, and this one is low in our list, sorry.

Issue #340, opened 2021-11-24
Drop support for Python 2


ccordoba12:
- I think we should drop support for Python 2 in our next version because it's making development too cumbersome.
- Besides, it's been almost two years since Python 2 development stopped, so there's no need to keep supporting it longer.

@impact27, could you give us a hand with this one?

PR #355, opened 2021-12-24
PR: Print last line of Pdb code and some debugger fixes

### LINKED ISSUE #330: spydercustomize in debug stack
When debugging a file, a few internal frames are present in the debugging stack. This is present when using pdb commands like "where" or "up". The base of the problem is that `debugfile` doesn't call the call directly, but instead calls `runfile`. The stack is therefore:

```
bdb.py::run()
spydercustomize.py::runfile()
spydercustomize.py::exec_code()
myfile.py
```

The central two can be hidden with the `skip_hidden` command, but they can still leak in other ways. For example when using `where` with `skip_hidden` activated, you see: `[... skipping 2 hidden frame(s)]`

impact27:
When entering code in ipython, if the last line is an expression, it is printed after an "Out [i]: " prompt. This PR reproduces this behaviour in the debugger.

Fixes #330

### DIFF
--- a/spyder_kernels/comms/frontendcomm.py
+++ b/spyder_kernels/comms/frontendcomm.py
@@ -34,7 +34,7 @@ def get_free_port():
     return port
 
 
-def frontend_request(blocking=True, timeout=None):
+def frontend_request(blocking, timeout=None):
     """
     Send a request to the frontend.
 
--- a/spyder_kernels/comms/frontendcomm.py
+++ b/spyder_kernels/comms/frontendcomm.py
@@ -285,14 +285,15 @@ class WriteWrapper():
 
     def __call__(self, string):
         """Print warning once."""
+        if self._thread_id != threading.get_ident():
+            return self._write(string)
+
         # Don't print DeprecationWarning's because they unnecessarily pollute
         # the console.
         # Fixes spyder-ide/spyder#14928
         # Fixes spyder-ide/spyder-kernels#343
         if 'DeprecationWarning' not in string:
-            if (self._thread_id == threading.get_ident()
-                and not self._warning_shown
-                ):
+            if not self._warning_shown:
                 self._warning_shown = True
                 self._write(
                     "\nOutput from spyder call "
--- a/spyder_kernels/comms/frontendcomm.py
+++ b/spyder_kernels/comms/frontendcomm.py
@@ -261,8 +261,11 @@ class FrontendComm(CommBase):
         """Call the callback function for the remote call."""
         saved_stdout_write = sys.stdout.write
         saved_stderr_write = sys.stderr.write
-        sys.stdout.write = WriteWrapper(saved_stdout_write, call_name)
-        sys.stderr.write = WriteWrapper(saved_stderr_write, call_name)
+        thread_id = threading.get_ident()
+        sys.stdout.write = WriteWrapper(
+            saved_stdout_write, call_name, thread_id)
+        sys.stderr.write = WriteWrapper(
+            saved_stderr_write, call_name, thread_id)
         try:
             return super(FrontendComm, self)._remote_callback(
                 call_name, call_args, call_kwargs)
@@ -274,9 +277,10 @@ class FrontendComm(CommBase):
 class WriteWrapper():
     """Wrapper to warn user when text is printed."""
 
-    def __init__(self, write, name):
+    def __init__(self, write, name, thread_id):
         self._write = write
         self._name = name
+        self._thread_id = thread_id
         self._warning_shown = False
 
     def __call__(self, string):
@@ -286,7 +290,9 @@ class WriteWrapper():
         # Fixes spyder-ide/spyder#14928
         # Fixes spyder-ide/spyder-kernels#343
         if 'DeprecationWarning' not in string:
-            if not self._warning_shown:
+            if (self._thread_id == threading.get_ident()
+                and not self._warning_shown
+                ):
                 self._warning_shown = True
                 self._write(
                     "\nOutput from spyder call "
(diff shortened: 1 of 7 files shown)


spyder_kernels/customize/spyderpdb.py
```diff
@@ -175,6 +175,24 @@ def default(self, line):
                 if execute_events:
                      get_ipython().events.trigger('pre_execute')
 
+                code_ast = ast.parse(line)
+                # Modify ast code to capture the last expression
+                capture_last_expression = False
+                if (len(code_ast.body)
+                        and isinstance(code_ast.body[-1], ast.Expr)
+                        and line.rstrip()[-1] != ";"):
```
ccordoba12, line 183:
```suggestion
                if (
                    len(code_ast.body)
                    and isinstance(code_ast.body[-1], ast.Expr)
                    and line.rstrip()[-1] != ";"
                ):
```

spyder_kernels/customize/spyderpdb.py
```diff
@@ -195,46 +213,40 @@ def default(self, line):
...
-                    try:
-                        code = compile(line + '\n', '<stdin>', 'single')
-                    except SyntaxError:
-                        # Support multiline statments
-                        code = compile(line + '\n', '<stdin>', 'exec')
-                    exec(code, globals)
+                    # Cleanup
+                    code += [
+                        "del _spyderpdb_code",
+                        "del _spyderpdb_locals"]
```
ccordoba12, line 230:
```suggestion
                        "del _spyderpdb_locals"
                    ]
```

spyder_kernels/customize/spyderpdb.py
```diff
@@ -195,46 +213,40 @@ def default(self, line):
...
+                    # Parse the function
+                    fun_ast = ast.parse('\n'.join(code) + '\n')
+
+                    # Inject code_ast in the function before the locals update
+                    fun_ast.body[0].body = (
+                        fun_ast.body[0].body[:-1]  # The locals
+                        + code_ast.body  # Code to run
+                        + fun_ast.body[0].body[-1:])  # Locals update
```
ccordoba12, line 238:
```suggestion
                        + fun_ast.body[0].body[-1:]  # Locals update
                    )
```

ccordoba12 (commented):
Thanks @impact27 for your help with this! It's really nice to have this behavior in Pdb as well.

Besides the minor formatting suggestions I left for you, could you add a test for this? It could be here or in the corresponding PR on the Spyder repo.

Finally, I'm going to leave this for Spyder-kernels 2.3.0/Spyder 5.3.0 because it requires a new handler.

spyder_kernels/customize/spydercustomize.py
```diff
@@ -569,15 +606,17 @@ def runfile(filename=None, args=None, wdir=None, namespace=None,
...
-        if __umr__.has_cython:
-            # Cython files
-            with io.open(filename, encoding='utf-8') as f:
-                ipython_shell.run_cell_magic('cython', '', f.read())
-        else:
-            exec_code(file_code, filename, ns_globals, ns_locals,
-                      post_mortem=post_mortem)
-
-        sys.argv = ['']
+        try:
```
ccordoba12, line 609: Why is this closed in a `try` now?

spyder_kernels/customize/spydercustomize.py
```diff
@@ -589,6 +628,14 @@ def runfile(filename=None, args=None, wdir=None, namespace=None,
     builtins.spyder_runfile = runfile
 
 
+def normalise_filename(filename):
```
ccordoba12, line 631: Let's move this function to `spyder_kernels/customize/utils.py`.

spyder_kernels/customize/spyderpdb.py
```diff
@@ -782,43 +773,42 @@ def runcall(self, *args, **kwds):
...
 
-
-def enter_debugger(filename, continue_if_has_breakpoints, code_format):
-    """Enter debugger. Code format should be a format that accept filename."""
-    shell = get_ipython()
-    recursive = shell.is_debugging()
-    if recursive:
-        parent_debugger = shell.pdb_session
+    def enter_recursive_debugger(
+            self, code, filename, continue_if_has_breakpoints):
```
ccordoba12, line 777:
```suggestion
    def enter_recursive_debugger(self, code, filename,
                                continue_if_has_breakpoints):
```

Minor style issue.

ccordoba12 (commented):
@impact27, some minor comments for you, then this should be ready.

After merging this one, I think you could proceed to finish PR #340 (I didn't notice it still touches too much code around).

spyder_kernels/comms/frontendcomm.py
```diff
@@ -274,13 +277,17 @@ def _remote_callback(self, call_name, call_args, call_kwargs):
...
+    def __init__(self, write, name, thread_id):
         self._write = write
         self._name = name
+        self._thread_id = thread_id
         self._warning_shown = False
 
     def __call__(self, string):
         """Print warning once."""
+        if self._thread_id != threading.get_ident():
+            return self._write(string)
```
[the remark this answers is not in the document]
impact27: Not even the main thread? If I put that inside `if 'DeprecationWarning' not in string:` no DeprecationWarning will be printed at all, even from the main thread.

spyder_kernels/customize/spydercustomize.py
impact27: Because in case of failure `sys.argv` is not reset. This was an undiscovered but that I am fixing here.

ccordoba12 (commented):
@impact27, I left some minor style suggestions for you, then this should be ready.

ccordoba12 (approved):
Thanks @impact27! Great work here.

PR #363, opened 2022-01-10
PR: Fix error when running multiprocessing code that contains classes

impact27:
See https://github.com/spyder-ide/spyder/issues/16696


ccordoba12 (approved):

PR #369, opened 2022-01-14
PR: Leave unsaved files in linecache

impact27:
Address https://github.com/spyder-ide/spyder/issues/16417 for unsaved files


ccordoba12 (approved):
Thanks @impact27!

PR #370, opened 2022-01-22
PR: Fix issues when debugging comprehensions

### LINKED ISSUE #371: Can't debug comprehensions
Executing code in a comprehension with the debugger results in a Syntax Error. Solved by https://github.com/spyder-ide/spyder-kernels/pull/370

impact27:
Namespace is broken in comprehensions because the iterator variable is saved as ".0", which is not a correct identifier name

- Test in https://github.com/spyder-ide/spyder/pull/17231
- Fixes #371


ccordoba12 (approved):
Thanks @impact27!

PR #376, opened 2022-03-07
PR: Add new handler to detect the current Matplotlib interactive backend

ccordoba12:
This is necessary to correctly inform users in Spyder that a kernel restart is necessary when they set first the Qt5 backend, then Inline, then Tk.

To account for that possibility, we need a way to detect the current interactive backend (e.g. Qt5 or Tk) independently of the current one. Before we were assuming that if the current backend is Inline (as in the second step above), then you can switch to any interactive one, but that's actually not correct because it's not possible to have two interactive backends running at the same time.

### DIFF
--- a/.github/workflows/linux-tests.yml
+++ b/.github/workflows/linux-tests.yml
@@ -18,6 +18,7 @@ jobs:
       CI: True
       PYTHON_VERSION: ${{ matrix.PYTHON_VERSION }}
       RUNNER_OS: 'ubuntu'
+      USE_CONDA: 'true'
     strategy:
       fail-fast: false 
       matrix:
--- a/spyder_kernels/console/kernel.py
+++ b/spyder_kernels/console/kernel.py
@@ -768,7 +768,12 @@ class SpyderKernel(IPythonKernel):
         """
         import traceback
         from IPython.core.getipython import get_ipython
-        import matplotlib
+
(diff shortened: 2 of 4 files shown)


dalthviz (commented):
Thanks @ccordoba12 ! Left a comment regarding an unhandled exception. Otherwise LGTM :+1:

dalthviz:
Checking this with Spyder if you try to change the backend without `matplotlib` installed you get a traceback in the console:

```python
  File "C:\Users\dalth\AppData\Local\Temp/ipykernel_4416/3542597582.py", line 1, in <module>
    get_ipython().kernel.set_matplotlib_backend('2', False)

  File "E:\Acer\Documentos\Spyder\Spyder otros\carlos\spyder\external-deps\spyder-kernels\spyder_kernels\console\kernel.py", line 509, in set_matplotlib_backend
    self._set_mpl_backend(mpl_backend, pylab=pylab)

  File "E:\Acer\Documentos\Spyder\Spyder otros\carlos\spyder\external-deps\spyder-kernels\spyder_kernels\console\kernel.py", line 771, in _set_mpl_backend
    import matplotlib

ModuleNotFoundError: No module named 'matplotlib'
```

imagen


Is that the expected behavior?

ccordoba12:
> Is that the expected behavior?

No, it isn't, and great catch by the way! I fixed it in my last commit (i.e. 6ce8770).

PR #377, opened 2022-03-08
PR: Handle getting dasks objects size

dalthviz:
See spyder-ide/spyder#16844

### DIFF
--- a/spyder_kernels/utils/nsview.py
+++ b/spyder_kernels/utils/nsview.py
@@ -85,15 +85,19 @@ def try_to_eval(value):
 def get_size(item):
     """Return shape/size/len of an item of arbitrary type"""
     try:
-        if (hasattr(item, 'size') and hasattr(item.size, 'compute') or
-                hasattr(item, 'shape') and hasattr(item.shape, 'compute')):
+        if (
+            hasattr(item, 'size') and hasattr(item.size, 'compute') or
+            hasattr(item, 'shape') and hasattr(item.shape, 'compute')
+        ):
             # This is necessary to avoid an error when trying to
             # get the size/shape of dask objects. We don't compute the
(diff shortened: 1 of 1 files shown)


spyder_kernels/utils/nsview.py
```diff
@@ -85,7 +85,14 @@ def try_to_eval(value):
 def get_size(item):
     """Return shape/size/len of an item of arbitrary type"""
     try:
-        if (hasattr(item, 'shape') and
+        if (hasattr(item, 'size') and hasattr(item.size, 'compute') or
+                hasattr(item, 'shape') and hasattr(item.shape, 'compute')):
```
ccordoba12, line 89:
```suggestion
        if (
            hasattr(item, 'size') and hasattr(item.size, 'compute') or
            hasattr(item, 'shape') and hasattr(item.shape, 'compute')
        ):
```

spyder_kernels/utils/nsview.py
```diff
@@ -85,7 +85,14 @@ def try_to_eval(value):
...
+            # This is necessary to avoid an error when trying to
+            # get the size/shape of dask objects. We don't compute the
+            # size/shape since such operation could be expensive.
+            # Fixes spyder-ide/spyder#16844
+            return 1
+        elif (hasattr(item, 'shape') and
                 isinstance(item.shape, (tuple, np.integer))):
```
ccordoba12, line 96:
```suggestion
        elif (
            hasattr(item, 'shape') and
            isinstance(item.shape, (tuple, np.integer))
        ):
```

ccordoba12 (commented):
Thanks @dalthviz, I left a couple of small suggestions for readability. Otherwise, looks good to me.

PR #381, opened 2022-05-03
PR: Eliminate unnecessary updates of the Variable Explorer while debugging

rear1019:
See commit message for a detailed description. We have been using this patch for quite some time and haven’t experienced issues so far.

At least in theory, this should improve the performance of the debugger in Spyder (in terms of steps/second). In practical terms, performance gains are only noticeable with _very_ many variables in current scope. Think of about 80–100 variables on a rather old/slow machine, and > 200–300 variables on a rather fast one. Gains may be bigger with remote kernels, but I haven’t tested that

PS: Please let me know if I should provide instructions how to (roughly) measure the performance.

### DIFF
--- a/spyder_kernels/console/kernel.py
+++ b/spyder_kernels/console/kernel.py
@@ -311,11 +311,7 @@ class SpyderKernel(IPythonKernel):
         Publish Variable Explorer state and Pdb step through
         send_spyder_msg.
         """
-        state = dict(
-            namespace_view=self.get_namespace_view(),
-            var_properties=self.get_var_properties(),
-            step=step
-         )
+        state = dict(step=step)
         try:
             self.frontend_call(blocking=False).pdb_state(state)
         except (CommError, TimeoutError):
--- a/spyder_kernels/console/kernel.py
+++ b/spyder_kernels/console/kernel.py
@@ -306,17 +306,6 @@ class SpyderKernel(IPythonKernel):
             return self.shell.pdb_session.do_complete(code, cursor_pos)
         return self._do_complete(code, cursor_pos)
 
-    def publish_pdb_state(self, step):
-        """
-        Publish Variable Explorer state and Pdb step through
-        send_spyder_msg.
-        """
-        state = dict(step=step)
-        try:
-            self.frontend_call(blocking=False).pdb_state(state)
-        except (CommError, TimeoutError):
-            logger.debug("Could not send Pdb state to the frontend.")
-
     def set_spyder_breakpoints(self, breakpoints):
         """
         Handle a message from the frontend
--- a/spyder_kernels/customize/spyderpdb.py
+++ b/spyder_kernels/customize/spyderpdb.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 #
 # Copyright (c) 2009- Spyder Kernels Contributors
 #
@@ -91,6 +92,9 @@ class SpyderPdb(ipyPdb, object):  # Inherits `object` to call super() in PY2
         super(SpyderPdb, self).__init__()
         self._pdb_breaking = False
         self._frontend_notified = False
+
+        # content of tuple: (filename, line number)
+        self._previous_step = None
 
         # Don't report hidden frames for IPython 7.24+. This attribute
         # has no effect in previous versions.
@@ -542,6 +546,12 @@ class SpyderPdb(ipyPdb, object):  # Inherits `object` to call super() in PY2
         return result
 
     # --- Methods overriden by us for Spyder integration
+    def postloop(self):
+        # postloop() is called when the debugger’s input prompt exists. Reset
+        # _previous_step so that publish_pdb_state() actually notifies Spyder
+        # about a changed frame the next the input prompt is entered again.
+        self._previous_step = None
+
     def preloop(self):
         """Ask Spyder for breakpoints before the first prompt is created."""
         try:
@@ -554,7 +564,7 @@ class SpyderPdb(ipyPdb, object):  # Inherits `object` to call super() in PY2
             if self.starting:
                 self.set_spyder_breakpoints(pdb_settings['breakpoints'])
             if self.send_initial_notification:
-                self.notify_spyder()
+                self.publish_pdb_state()
         except (CommError, TimeoutError):
             logger.debug("Could not get breakpoints from the frontend.")
         super(SpyderPdb, self).preloop()
@@ -644,14 +654,11 @@ class SpyderPdb(ipyPdb, object):  # Inherits `object` to call super() in PY2
 
     def postcmd(self, stop, line):
         """
-        Notify spyder about a changed frame
+        Notify spyder about (possibly) changed frame
+
+        Note: The PDB commands “up”, “down” and “jump” change the current frame.
         """
-        line = line.strip()
-        if line.startswith("!"):
-            line = line[1:]
-        cmd, arg, line = self.parseline(line.strip())
-        if cmd in ("up", "down", "jump", "u", "d", "j"):
-            self.notify_spyder()
+        self.publish_pdb_state()
         return super(SpyderPdb, self).postcmd(stop, line)
 
     if PY2:
@@ -748,11 +755,16 @@ class SpyderPdb(ipyPdb, object):  # Inherits `object` to call super() in PY2
                     logger.debug(
                         "Could not send a Pdb continue call to the frontend.")
 
-    def notify_spyder(self):
-        """Send kernel state to the frontend."""
+    def publish_pdb_state(self):
+        """
+        Send debugger state (frame position) to the frontend.
+
+        The state is only sent if it has changed since the last update.
+        """
 
         frame = self.curframe
         if frame is None:
+            self._previous_step = None
             return
 
         # Get filename and line number of the current frame
@@ -764,12 +776,20 @@ class SpyderPdb(ipyPdb, object):  # Inherits `object` to call super() in PY2
                 pass
         lineno = frame.f_lineno
 
+        if self._previous_step == (fname, lineno):
+            return
+
         # Set step of the current frame (if any)
         step = {}
+        self._previous_step = None
         if isinstance(fname, basestring) and isinstance(lineno, int):
             step = dict(fname=fname, lineno=lineno)
+            self._previous_step = (fname, lineno)
 
-        get_ipython().kernel.publish_pdb_state(step)
+        try:
+            frontend_request(blocking=False).pdb_state(dict(step=step))
+        except (CommError, TimeoutError):
+            logger.debug("Could not send Pdb state to the frontend.")
 
     def run(self, cmd, globals=None, locals=None):
         """Debug a statement executed via the exec() function.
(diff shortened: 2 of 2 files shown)


ccordoba12:
@impact27, could you take a look at this one? Thanks!

impact27:
Is `update_plugins` ever set to True? I can't see any instances in the PR.
If you add a variable with a command (simply `testvar = 1` in the debugger), is it shown in the variable explorer?

rear1019:
> Is `update_plugins` ever set to True? I can't see any instances in the PR.

This is not required (also see below). I have added this parameter without a default value so that any user of `publish_pdb_state()` or `notify_spyder()` must explicitly check whether `update_plugins` must be `True` or `False`. However, I do see that this is confusing, so it’s possibly better to remove this parameter.

> If you add a variable with a command (simply `testvar = 1` in the debugger), is it shown in the variable explorer?

Yes, it is shown. Any modification of the current namespace is shown, just like before. I have tried to explain this patch in the lengthy commit message, but let’s give it another try:

The are two ways the Variables Explorer is updated:
* The kernel proactively sends updated data. This only happens inside the debugger.
* The Variable Explorer requests updated data from the kernel.

The former is not required due to the latter and is removed by this PR. The latter relies on the `executed` signal emitted by `QtConsole` [1]: The Variable Explorer requests an update of its data from the kernel upon the `executed` signal [2].

This signal is emitted when the kernel finishes execution of a command (in the non-debugger case). As an extension, this signal is also emitted by Spyder when the debugger requests input. The request is a remote call into Spyder, handled in `DebuggingWidget.pdb_input()` [2]. In any case, the `executed` signal signals that the kernel has finished execution of a command, implying that the namespace may have changed.

[1] https://github.com/jupyter/qtconsole/blob/5.3.0/qtconsole/frontend_widget.py#L145-L147
[2] https://github.com/spyder-ide/spyder/blob/v5.3.0/spyder/plugins/ipythonconsole/widgets/client.py#L427
[3] https://github.com/spyder-ide/spyder/blob/v5.3.0/spyder/plugins/ipythonconsole/widgets/debugging.py#L599

impact27:
> The are two ways the Variables Explorer is updated:
> 
>     * The kernel proactively sends updated data. This only happens inside the debugger.
> 
>     * The Variable Explorer requests updated data from the kernel.
> 
> The former is not required due to the latter and is removed by this PR. 

This makes sense. Is there a reason why you elected to remove the former rather than the latter? Removing the latter instead would avoid an extra request wouldn't it?

Also the `executed` signal is only emitted when the execution is visible[1]. If Spyder does "hidden execution" there would be no request to update the variable explorer and it would therefore not be up to date? (To be fair I am not sure under which circumstances that would happen, maybe when updating a variable from the variable explorer?)

> This is not required (also see below). I have added this parameter without a default value so that any user of publish_pdb_state() or notify_spyder() must explicitly check whether update_plugins must be True or False. However, I do see that this is confusing, so it’s possibly better to remove this parameter.

I don't think it is needed to keep that if it is not used.


[1]https://github.com/spyder-ide/spyder/blob/c4cca83a68f9bd69ff2afb9d4396ea3d297bfdbc/spyder/plugins/ipythonconsole/widgets/debugging.py#L623

rear1019:
> > The are two ways the Variables Explorer is updated:
> > ```
> > * The kernel proactively sends updated data. This only happens inside the debugger.
> > 
> > * The Variable Explorer requests updated data from the kernel.
> > ```
> > The former is not required due to the latter and is removed by this PR.
> 
> This makes sense. Is there a reason why you elected to remove the former rather than the latter? Removing the latter instead would avoid an extra request wouldn't it?

Yes, removing the latter (update requests by Spyder) and retaining the former (kernel proactively sends data) would avoid an extra request. This is something we considered, and we had a corresponding patch for the debugger-case (IIRC). IIRC, we opted not go this way because
* We already had a working patch (this PR).
* We didn’t have the time to find and verify a corresponding patch for the non-debugger case (“plain” IPython console).
* We didn’t want to fix only the debugger-case, leading a situation where the debugger and non-debugger case use different mechanisms. 

However, making this further improvement – kernel proactively sends data instead of Spyder requesting it – is still in our backlog :wink:.

> 
> Also the `executed` signal is only emitted when the execution is visible[1].
>
> [1]https://github.com/spyder-ide/spyder/blob/c4cca83a68f9bd69ff2afb9d4396ea3d297bfdbc/spyder/plugins/ipythonconsole/widgets/debugging.py#L623

Note: This also applies to the non-debugger case (“plain” IPython console) [1]

> If Spyder does "hidden execution" there would be no request to update the variable explorer and it would therefore not be up to date?

That’s correct, assuming the hidden execution modifies the current namespace (I hope not). However, this issue exists regardless whether this PR is applied or not.

To be honest, I didn’t looked too hard into hidden execution. I merely verified that debugger commands triggered by Spyder’s GUI are not hidden.

> (To be fair I am not sure under which circumstances that would happen, maybe when updating a variable from the variable explorer?)
> 

Changing a variable from within the Variable Explorer (as opposed to entering code in the IPython console) uses a different communication channel implemented by `spyder_kernels`/Spyder. No execution requests/replies (as defined by Jupyter and implemented by `ipykernel/qtconsole`) are involved and consequently no `executed` signal is emitted. That’s actually an issue for the Watchlist plugin I wasn’t aware of.

> > This is not required (also see below). I have added this parameter without a default value so that any user of publish_pdb_state() or notify_spyder() must explicitly check whether update_plugins must be True or False. However, I do see that this is confusing, so it’s possibly better to remove this parameter.
> 
> I don't think it is needed to keep that if it is not used.

I will remove it.

[1] https://github.com/jupyter/qtconsole/blob/5.3.0/qtconsole/frontend_widget.py#L449

impact27:
> Note: This also applies to the non-debugger case (“plain” IPython console) [1]

This is a good point. I think we should keep the same code path for both debugger and non-debugger then.

I tried this PR and it breaks debugging following in spyder. If you use the up/down command the blue arrow doesn not follow the position of the debugger. I think you need to add back `postcmd`.

To test:

```python
test1 = 1
def a():
    test2 = 2
    return 1 + 1 # Breakpoint here
a()
```

1. debug file
2. use up command
3. See that the debug arrow is still on line 4

rear1019:
Changes in force push:
* Fixed issue found by @impact27 [1]
* Removed `update_plugins` parameter

[1] https://github.com/spyder-ide/spyder-kernels/pull/381#issuecomment-1129987411

spyder_kernels/customize/spyderpdb.py
```diff
@@ -644,9 +644,14 @@ def precmd(self, line):
...
         """
-        Notify spyder on any pdb command.
+        Notify spyder about a changed frame
         """
-        self.notify_spyder()
+        line = line.strip()
+        if line.startswith("!"):
+            line = line[1:]
+        cmd, arg, line = self.parseline(line.strip())
+        if cmd in ("up", "down", "jump", "u", "d", "j"):
```
impact27, line 653: It is a good idea to check wether the update is needed!

I am not sure I am a fan of the list of commands. This duplicates a part of the `line` processing and the list might be subject to change. Why not check if the frame has changed instead, which is what we want to know?

My proposition would be to move the logic inside `notify_spyder`, where the filename and line number can be compared to the previous update, saved in a variable (`self._previous_step` maybe?).
Then `self._previous_step = None` in `postloop` so the next `preloop` would send an update as expected.

Also maybe `notify_spyder` could be renamed now that it only sends the debugger position?

impact27 (changes requested):
Thanks for the work, I think this will help with the debugging speed!

spyder_kernels/console/kernel.py
```diff
@@ -311,11 +311,7 @@ def publish_pdb_state(self, step):
         Publish Variable Explorer state and Pdb step through
         send_spyder_msg.
         """
-        state = dict(
-            namespace_view=self.get_namespace_view(),
-            var_properties=self.get_var_properties(),
-            step=step
-         )
+        state = dict(step=step)
```
impact27, line 314: Now that `publish_pdb_state` doesn't use any kernel function I don't think it should stay in kernel. I would move the logic here in `notify_spyder`. 

You can replace `self.frontend_call(blocking=False)` by `frontend_request(blocking=False)`

rear1019:
Thanks for the review @impact27! I have implemented your very sensible suggestions. Please let me now if I should rebase the patch against the most recent `master` or `2.x` branch.

impact27:
Thanks! Looks good now!

@ccordoba12 what do you think about the branch?

spyder_kernels/customize/spyderpdb.py
```diff
@@ -92,6 +93,9 @@ def __init__(self, completekey='tab', stdin=None, stdout=None,
         self._pdb_breaking = False
         self._frontend_notified = False
 
+        # content of tuple: (filename, line number)
+        self._previous_step: Optional[Tuple[basestring, int]] = None
```
ccordoba12, line 97:
```suggestion
        self._previous_step = None
```

spyder_kernels/customize/spyderpdb.py
```diff
@@ -13,6 +13,7 @@
 import sys
 import traceback
 from collections import namedtuple
+from typing import Tuple, Optional
```
ccordoba12, line 16:
```suggestion
```

This package still supports Python 2, so we need to remove this and the typing annotations you added below. We will migrate to Python 3 only in Spyder-kernels 3.

spyder_kernels/customize/spyderpdb.py
```diff
@@ -542,6 +546,12 @@ def is_name_or_composed(text):
         return result
 
     # --- Methods overriden by us for Spyder integration
+    def postloop(self) -> None:
```
ccordoba12, line 549:
```suggestion
    def postloop(self):
```

spyder_kernels/customize/spyderpdb.py
```diff
@@ -743,11 +754,13 @@ def set_spyder_breakpoints(self, breakpoints):
                     logger.debug(
                         "Could not send a Pdb continue call to the frontend.")
 
-    def notify_spyder(self):
-        """Send kernel state to the frontend."""
+    def publish_pdb_state(self):
+        """Send debugger state (frame position) to the frontend. The state is
+        only sent if it has changed since the last update."""
```
ccordoba12, line 759:
```suggestion
        """
        Send debugger state (frame position) to the frontend.

        The state is only sent if it has changed since the last update.
        """
```

Improve formatting of this docstring.

ccordoba12 (commented):
@rear1019, last review for you, then this should be ready.

rear1019:
Force push:
* Fixed display of current code position (blue arrow in editor) after leaving a recursive debugger. @impact27 See the changes in `SpyderPdb.enter_recursive_debugger()`.
* Applied suggestions by @ccordoba12 to restore compatibility with Python 2
* Rebased against current `2.x` branch

impact27:
> * See the changes in `SpyderPdb.enter_recursive_debugger()`.

Looks good thanks for checking that

spyder_kernels/customize/spyderpdb.py
```diff
@@ -821,6 +845,11 @@ def enter_recursive_debugger(self, code, filename,
         self.lastcmd = debugger.lastcmd
         get_ipython().pdb_session = self
 
+        # Notify spyder about changed frame. Reset _previous_step to force
+        # notification.
+        self._previous_step = None
+        self.publish_pdb_state()
```
impact27, line 851: I think this is not needed because `postcmd` will be called after this function returns. The line above (`self._previous_step = None`) should be sufficient to address the issue. Am I missing something?

rear1019: No, you’re not. Resetting `self._previous_step` is sufficient.

rear1019:
Force push:
* Addressed comment [1]
* Added encoding line in `spyderpdb.py`. This should fix CI failures in with Python 2

[1] https://github.com/spyder-ide/spyder-kernels/pull/381#discussion_r897941057

ccordoba12:
Thanks @rear1019 for the updates!

@impact27, what do you think about them?

impact27:
Looks good to me 👍

ccordoba12:
Great! Merging then.

ccordoba12 (approved):
Thanks @rear1019 for your work on this!

PR #383, opened 2022-05-19
PR: Add required code for external Watchlist plugin

rear1019:
This is a draft PR so that the required changes for the external Watchlist plugin for Spyder [1] can be reviewed. Please note that this PR also contains the changes from PR #381. They will be removed from this PR once PR #381 is merged.

[1] https://github.com/spyder-ide/spyder/issues/16438


ccordoba12:
@rear1019, please rebase and finish this one so we can include it in our next release of Spyder-kernels (2.3.2)

rear1019:
> @rear1019, please rebase and finish this one so we can include it in our next release of Spyder-kernels (2.3.2)

Done

impact27:
I am not sure external plugins should modify the spyder-kernels. I tried to think of a solution and came up with https://github.com/spyder-ide/spyder-kernels/pull/392

The idea is that you would inject the code during setup. 
For this PR the setup code would look like:
```
"""
kernel = get_ipython().kernel
kernel.watchlist_expressions = []
kernel.watchlist_debugger_only = True
"""
```
Then the comm handlers would be:
```
{"set_watchlist_expressions": """
def set_watchlist_expressions(expressions, debugger_only=True):
    kernel = get_ipython().kernel
    kernel.watchlist_expressions = expressions
    kernel.watchlist_debugger_only = debugger_only
""",
"eval_watchlist_expressions": """
def eval_watchlist_expressions():
    ...
"""
}
```
What do you think of this approach?

rear1019:
> I am not sure external plugins should modify the spyder-kernels.

This is something I am not quite content with as well. Mostly because it makes it harder to update (fix) code in `spyder-kernels` for external plugins.

> I tried to think of a solution and came up with https://github.com/spyder-ide/spyder-kernels/pull/392
>
> The idea is that you would inject the code during setup.
> …
> What do you think of this approach?

It works for the Watchlist and there is an other use case as well (spyder-ide/spyder-unittest#177). One might consider this approach somewhat hacky, but I can’t think of a better way.

ccordoba12:
@rear1019, the problem is that PR #392 is targeted for Spyder-kernels 3.0, which will be released alongside Spyder 6.0 next year.

I'm fine with including your changes in Spyder-kernels 2.3.2 (our next version) so you can start using the Watchlist plugin now. Then, you could move them to your plugin when Spyder 6 is released. What do you think?

impact27:
Alternatively you can simply execute in the console something like:
```
from IPython.core.getipython import get_ipython

kernel = get_ipython().kernel
kernel.watchlist_expressions = []
kernel.watchlist_debugger_only = True

def set_watchlist_expressions(expressions, debugger_only=True):
    kernel.watchlist_expressions = expressions
    kernel.watchlist_debugger_only = debugger_only

kernel.frontend_comm.register_call_handler(
                "set_watchlist_expressions", set_watchlist_expressions)
```
using `shellwidget.execute(source, hidden=True)`

impact27:
If you don't want to pollute the namespace:
```python
def register_plugin():
    from IPython.core.getipython import get_ipython
    
    kernel = get_ipython().kernel
    kernel.watchlist_expressions = []
    kernel.watchlist_debugger_only = True
    
    def set_watchlist_expressions(expressions, debugger_only=True):
        kernel.watchlist_expressions = expressions
        kernel.watchlist_debugger_only = debugger_only
    
    kernel.frontend_comm.register_call_handler(
                    "set_watchlist_expressions", set_watchlist_expressions)

register_plugin()
del register_plugin
```

rear1019:
For now I think it’s best to go with the suggestion by @impact27. It works and is more flexible than relying on upstream changes in `spyder-kernels`.

Thank you very much for your support and valuable input, @ccordoba12 and @impact27!

PR #389, opened 2022-06-20
PR: Fix debug filename path for remote debugging

impact27:
On linux when processing window paths:
```
IPdb [5]: os.path.abspath("C:/Users")
Out  [5]: '/home/debian/spyder-kernels/C:/Users'
```
This PR changes the filename handling to fix https://github.com/spyder-ide/spyder/issues/18330

### DIFF
--- a/spyder_kernels/customize/spydercustomize.py
+++ b/spyder_kernels/customize/spydercustomize.py
@@ -31,7 +31,7 @@ from spyder_kernels.customize.spyderpdb import SpyderPdb, get_new_debugger
 from spyder_kernels.customize.umr import UserModuleReloader
 from spyder_kernels.py3compat import (
     PY2, _print, encode, compat_exec, FileNotFoundError)
-from spyder_kernels.customize.utils import capture_last_Expr
+from spyder_kernels.customize.utils import capture_last_Expr, canonic
 
 if not PY2:
     from IPython.core.inputtransformer2 import (
@@ -535,20 +535,6 @@ def runfile(filename=None, args=None, wdir=None, namespace=None,
     return _exec_file(
         filename, args, wdir, namespace,
         post_mortem, current_namespace, stack_depth=1)
-
-
-def canonic(filename):
-    """
-    Return canonical form of filename.
-
-    This is a copy of bdb.canonic, so that the debugger will process 
-    filenames in the same way
-    """
-    if filename == "<" + filename[1:-1] + ">":
-        return filename
-    canonic = os.path.abspath(filename)
-    canonic = os.path.normcase(canonic)
-    return canonic
 
 
 def _exec_file(filename=None, args=None, wdir=None, namespace=None,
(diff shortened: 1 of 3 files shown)


spyder_kernels/customize/spydercustomize.py
```diff
@@ -536,19 +540,14 @@ def runfile(filename=None, args=None, wdir=None, namespace=None,
...
-               exec_fun=None):
+               exec_fun=None, canonic_filename=None):
     # Tell IPython to hide this frame (>7.16)
     __tracebackhide__ = True
     ipython_shell = get_ipython()
     if filename is None:
         filename = get_current_file_name()
         if filename is None:
             return
-    else:
```
impact27, line 547: not needed because remote filename and canonic filename are separated

spyder_kernels/customize/spydercustomize.py
```diff
@@ -655,7 +657,7 @@ def debugfile(filename=None, args=None, wdir=None, post_mortem=False,
     if shell.is_debugging():
         # Recursive
         code = (
-            "runfile({}".format(repr(normalise_filename(filename))) +
```
impact27, line 658: removed normalise_filename because filename is now the remote one

impact27:
I think the `debugger._user_requested_quit = False` line in `spyderpdb.py` is useless and can be removed

ccordoba12 (commented):
Thanks for your help with this @impact27! Just two tiny suggestions for you.

ccordoba12:
@impact27, how confident do you feel about these changes for a bug fix release? The thing is I wouldn't like to break our debugger just to fix a bug for remote environments. That's important, of course, but maintaining our debugger's reliability is more important at this point in the Spyder 5 development cycle.

impact27:
> @impact27, how confident do you feel about these changes for a bug fix release? The thing is I wouldn't like to break our debugger just to fix a bug for remote environments. That's important, of course, but maintaining our debugger's reliability is more important at this point in the Spyder 5 development cycle.

I mean it touches file paths, so there might be some hard-to-understand problem. I am relatively confident but because it touches file paths I am not completely comfortable.

ccordoba12:
> I think the debugger._user_requested_quit = False line in spyderpdb.py is useless and can be removed

I missed this comment, sorry. It's up to you if you want to remove it as part of this PR.

impact27:
> I missed this comment, sorry. It's up to you if you want to remove it as part of this PR.

I searched for `_user_requested_quit`. The only place it is used is in `pdb.py::main()`. The relevant code is:
```python
if run_as_module:
    pdb._runmodule(mainpyfile)
else:
    pdb._runscript(mainpyfile)
if pdb._user_requested_quit:
    break
```
`_user_requested_quit` is set in both `_runmodule` and `_runscript`. I therefore not see the point in setting it ourself.

ccordoba12 (approved):
Thanks @impact27!

Issue #394, opened 2022-07-01
The variable explorer is broken while debugging


impact27:
To reproduce:
debug a file using containing (using `debugfile`):
```python
main_ns = True
print()
```
After the first step, the variable explorer shows `main_ns` as expected
After the second step, the debugger enters inside `print()`. The variable explorer shows the locals, as exoected, but still shows `main_ns` instead of the `iostream.py` globals

PR #395, opened 2022-07-01
PR: Fix running namespace and improve eventloop integration while debugging

impact27:
Fixes https://github.com/spyder-ide/spyder-kernels/issues/394

### DIFF
--- a/spyder_kernels/customize/spydercustomize.py
+++ b/spyder_kernels/customize/spydercustomize.py
@@ -13,7 +13,6 @@
 
 import ast
 import bdb
-import cmd
 import io
 import logging
 import os
@@ -328,17 +327,7 @@ os.get_terminal_size = _patched_get_terminal_size
 # =============================================================================
 # Pdb adjustments
 # =============================================================================
-def cmd_input(prompt=''):
-    return get_ipython().kernel.cmd_input(prompt)
-
-
 pdb.Pdb = SpyderPdb
-
-if PY2:
-    cmd.raw_input = cmd_input
-else:
-    cmd.input = cmd_input
-
 
 # =============================================================================
 # User module reloader
--- a/spyder_kernels/customize/spyderpdb.py
+++ b/spyder_kernels/customize/spyderpdb.py
@@ -100,6 +100,8 @@ class SpyderPdb(ipyPdb, object):  # Inherits `object` to call super() in PY2
         # has no effect in previous versions.
         self.report_skipped = False
 
+        self.executing = False
+
     # --- Methods overriden for code execution
     def print_exclamation_warning(self):
         """Print pdb warning for exclamation mark."""
@@ -642,6 +644,7 @@ class SpyderPdb(ipyPdb, object):  # Inherits `object` to call super() in PY2
 
         Here we switch ! and non !
         """
+        self.executing = True
         if not self.pdb_use_exclamation_mark:
             return line
         if not line:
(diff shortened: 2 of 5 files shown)


spyder_kernels/customize/spyderpdb.py
```diff
@@ -100,6 +100,8 @@ def __init__(self, completekey='tab', stdin=None, stdout=None,
         # has no effect in previous versions.
         self.report_skipped = False
 
+        self.executing = False
```
ccordoba12, line 103: Please add a comment on top of this variable to explain why it's necessary or how it's used.

ccordoba12 (commented):
@impact27, I left a tiny comment for you, otherwise this looks good to me.

spyder_kernels/customize/spyderpdb.py
impact27: Slight change: What I am actually interested in is to know if a pdb prompt is waiting. I adjusted the code and moved `cmd_input` which should have been in `SpyderPdb`

spyder_kernels/customize/spydercustomize.py
```diff
@@ -328,18 +327,8 @@ def _patched_get_terminal_size(fd=None):
...
-def cmd_input(prompt=''):
-    return get_ipython().kernel.cmd_input(prompt)
-
-
 pdb.Pdb = SpyderPdb
 
-if PY2:
-    cmd.raw_input = cmd_input
-else:
-    cmd.input = cmd_input
```
impact27, line 340: That was hacky, I am glad to have found a better way

spyder_kernels/customize/spyderpdb.py
```diff
@@ -99,7 +101,14 @@ def __init__(self, completekey='tab', stdin=None, stdout=None,
         # has no effect in previous versions.
         self.report_skipped = False
 
-        self.executing = False
+        # Keep track of remote filename
```
ccordoba12, line 104:
```suggestion

        # Keep track of remote filename
```

spyder_kernels/customize/spyderpdb.py
```diff
@@ -99,7 +101,14 @@ def __init__(self, completekey='tab', stdin=None, stdout=None,
...
+        self.remote_filename = None
+
+        # State of the prompt
+        self.prompt_waiting = False
+
+        # Line recieved from the frontend
```
ccordoba12, line 110:
```suggestion
        # Line received from the frontend
```

spyder_kernels/customize/spyderpdb.py
```diff
@@ -636,14 +645,82 @@ def _cmdloop(self):
                        "For copying text while debugging, use Ctrl+Shift+C",
                        file=self.stdout)
 
+    def cmdloop(self, intro=None):
+        """Repeatedly issue a prompt, accept input, parse an initial prefix
+        off the received input, and dispatch to action methods, passing them
+        the remainder of the line as argument.
+
+        """
+
```
ccordoba12, line 650:
```suggestion

    def cmdloop(self, intro=None):
        """
        Repeatedly issue a prompt, accept input, parse an initial prefix
        off the received input, and dispatch to action methods, passing them
        the remainder of the line as argument.
        """
```

ccordoba12:
This looks good to me now @impact27. As before, I'd like to ask you: do you think it's worth it to include these changes in 2.3.2 ? Considering stability in mind, of course.

impact27:
The bug is quite bad because the namespace is wrong while debigging outside of the main file. I think the risk is relatively low, because the tests now cover this part better.

ccordoba12 (approved):
Thanks @impact27!

PR #396, opened 2022-07-01
PR: Add handlers to interrupt executions and enter the debugger after that

impact27:
This PR adds two functions to the comms:
 - `request_pdb_stop` that enters the debugger immediately (at the next python interaction). If the kernel is not debugging, it creates a debugger.
 - `raise_interrupt_signal` that raises an interrupt signal on request. This allows interrupting remote kernels.

The `shell.pdb_session` handling is improved as before it would not be unregistered in some cases. Now it returns the tracing pdb session. It handles recursive debugger by keeping a stack of debuggers.

A new function handles interrupt signals: `shell.spyderkernel_sigint_handler`. It is reset at every prompt to keep the behaviour constant from ipykernel which resets the SIGINT handler at every message (in `post_handler_hook`). It will play nice with a currently debugging shell. This completely replaces `Pdb.sigint_handler` that is disabled.

Any frame in spyder-kernels is hidden to avoid problems when interrupting.

### DIFF
--- a/spyder_kernels/console/shell.py
+++ b/spyder_kernels/console/shell.py
@@ -12,15 +12,23 @@ Spyder shell for Jupyter kernels.
 
 # Standard library imports
 import bdb
+import logging
+import os
+import signal
 import sys
 import traceback
+from _thread import interrupt_main
 
 # Third-party imports
 from ipykernel.zmqshell import ZMQInteractiveShell
 
 # Local imports
+from spyder_kernels.customize.spyderpdb import SpyderPdb
 from spyder_kernels.comms.frontendcomm import CommError
 from spyder_kernels.utils.mpl import automatic_backend
+
+
+logger = logging.getLogger(__name__)
 
 
 class SpyderShell(ZMQInteractiveShell):
@@ -29,7 +37,9 @@ class SpyderShell(ZMQInteractiveShell):
     def __init__(self, *args, **kwargs):
         # Create _pdb_obj_stack before __init__
         self._pdb_obj_stack = []
+        self._request_pdb_stop = False
         super(SpyderShell, self).__init__(*args, **kwargs)
+        self.register_debugger_sigint()
 
     def _showtraceback(self, etype, evalue, stb):
         """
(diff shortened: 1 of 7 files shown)


spyder_kernels/console/tests/test_console_kernel.py
```diff
@@ -1157,5 +1239,103 @@ def check_found(msg):
...
+                # not from my request
+                continue
+            break
+        assert time.time() - t0 < 5
+
+
+def test_enter_debug():
+    """
+    Test that using `global` triggers a warning.
+    """
```
ccordoba12, line 1300:
```suggestion
def test_enter_debug_after_interruption():
    """
    Test that we can enter the debugger after interrupting the current
    execution.
    """
```

spyder_kernels/console/tests/test_console_kernel.py
```diff
@@ -1157,5 +1239,103 @@ def check_found(msg):
...
+        comm._send_channel = client.control_channel
+        kernel_comm._register_comm(comm)
+
+        client.execute_interactive("import time", timeout=TIMEOUT)
+
+        # Try interrupting loop
+        t0 = time.time()
+        msg_id = client.execute("for i in range(100): time.sleep(.1)")
+        time.sleep(.2)
+        # Raise interrupt on control_channel
```
ccordoba12, line 1319:
```suggestion
        # Request to enter the debugger
```

spyder_kernels/customize/spyderpdb.py
```diff
@@ -98,11 +96,18 @@ def __init__(self, completekey='tab', stdin=None, stdout=None,
         self.remote_filename = None
 
         # State of the prompt
+        # Needed to know which namespace to show (user or current frame)
         self.prompt_waiting = False
 
         # Line received from the frontend
         self._cmd_input_line = None
 
+        # Disable sigint so we can do it ourself
```
ccordoba12, line 105:
```suggestion
        # Disable sigint so we can do it ourselves
```

ccordoba12 (commented):
Thanks @impact27 for your work on this! It's a really impressive improvement!!

My suggestions are simple style issues.

spyder_kernels/console/shell.py
```diff
@@ -77,12 +88,33 @@ def is_debugging(self):
...
+            return
+        self._pdb_obj_stack.append(pdb_obj)
+        try:
+            self.kernel.frontend_call(blocking=True).set_debug_state(
+                len(self._pdb_obj_stack))
+        except (CommError, TimeoutError):
+            logger.debug("Could not send debugging state to the frontend.")
+
+    def remove_pdb_session(self, pdb_obj):
+        """Remove a pdb object to the stack."""
```
ccordoba12, line 108:
```suggestion
        """Remove a pdb object from the stack."""
```

ccordoba12 (commented):
@impact27, last tiny suggestion then this should be ready.

ccordoba12 (approved):
Thanks @impact27!

PR #398, opened 2022-07-04
PR: Fix module namespace

impact27:
Avoid clearing the cached namespace in case of recursive call to runfile/debugfile. For example, calling `debugfile(...)` on a file and then at the pdb prompt running `runfile(...)` on the same file


ccordoba12 (approved):
Thanks @impact27!

PR #401, opened 2022-07-12
PR: Wait for connection file to be written

impact27:
Sometimes, the connection file is not yet written when the client tries to load it. This remove test flakyness.


ccordoba12 (approved):

PR #403, opened 2022-07-27
PR: Make debugger faster by avoiding unnecessary comm messages

impact27:
Faster pdb by decreasing the number of comm messages

### DIFF
--- a/spyder_kernels/console/kernel.py
+++ b/spyder_kernels/console/kernel.py
@@ -82,7 +82,7 @@ class SpyderKernel(IPythonKernel):
             'is_special_kernel_valid': self.is_special_kernel_valid,
             'get_matplotlib_backend': self.get_matplotlib_backend,
             'get_mpl_interactive_backend': self.get_mpl_interactive_backend,
-            'pdb_input_reply': self.pdb_input_reply,
+            'pdb_input_reply': self.shell.pdb_input_reply,
             'enable_faulthandler': self.enable_faulthandler,
             "flush_std": self.flush_std,
             'get_current_frames': self.get_current_frames,
@@ -102,7 +102,6 @@ class SpyderKernel(IPythonKernel):
         self.control_handlers['comm_msg'] = self.control_comm_msg
         self.control_handlers['complete_request'] = self.shell_handlers[
             'complete_request']
-        self.shell_handlers["interrupt_eventloop"] = self._interrupt_eventloop
 
         # Socket to signal shell_stream locally
         self.loopback_socket = None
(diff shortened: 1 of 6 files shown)


ccordoba12 (commented):
Thanks @impact27 for your work on this! I left some simple suggestions for you, otherwise looks good to me.

spyder_kernels/console/kernel.py
```diff
@@ -95,10 +95,24 @@ def __init__(self, *args, **kwargs):
...
+
+    def record_ports(self, ports):
+        """Record ports used by shell."""
+        super().record_ports(ports)
+        # Add socket to signal shell_stream locally
+        self.shell_handlers["interrupt_eventloop"] = self._interrupt_eventloop
+        self.loopback_socket = self.shell_stream.socket.context.socket(
+            zmq.DEALER)
+        self.loopback_socket.connect(
+            "tcp://127.0.0.1:%i" % self._recorded_ports['shell_port'])
```
ccordoba12, line 115: I'm not so sure about this: is it really necessary to alter the Jupyter protocol like this to interrupt the eventloop? If possible, I'd really prefer to keep the current comm for this because it's simpler and it can't be broken by changes in jupyter_client, for instance.

impact27: Well Currently you need to send a message to the frontend that will then send a message back to the kernel just to interrupt the event loop. If you have a remote kernel that is quite steep. What I do here is quite straightforward, as I am simply connecting to the shell port. If the call to _recorded_ports is the issue I could read the port from the connection file

impact27: I tried to address your concerns in the latest commit

spyder_kernels/console/shell.py
```diff
@@ -206,20 +248,44 @@ def request_pdb_stop(self):
...
-        pdb_session = self.pdb_session
         if self._request_pdb_stop:
             # SIGINT called from request_pdb_stop
             self._request_pdb_stop = False
             debugger = SpyderPdb()
             debugger.interrupt()
             debugger.set_trace(frame)
-        elif pdb_session:
+            return
+        pdb_session = self.pdb_session
```
ccordoba12, line 258:
```suggestion

        pdb_session = self.pdb_session
```

ccordoba12 (commented):
@impact27, this is another small review for you (I don't know if this is ready, but it seems close to be).

impact27:
I think this is ready now

ccordoba12:
Great! I don't have more comments for you, so please update PR https://github.com/spyder-ide/spyder/pull/18837 to see if tests pass on the Spyder side and then I'll merge this one.

ccordoba12 (approved):

PR #404, opened 2022-07-30
PR: Use debugging namespace when curframe is active

### LINKED ISSUE #405: Python 2 tests are broken in Conda slots
```
  File "/usr/share/miniconda3/envs/test/bin/pytest", line 11, in <module>
    sys.exit(main())
TypeError: 'NoneType' object is not callable
```

impact27:
- Fixes #405.
- Addresses https://github.com/spyder-ide/spyder/issues/18856

Also remove unnecessary del (function returns which leads to the ref being deleted anyway)

### DIFF
--- a/spyder_kernels/console/tests/test_console_kernel.py
+++ b/spyder_kernels/console/tests/test_console_kernel.py
@@ -1138,6 +1138,9 @@ def test_get_interactive_backend(backend):
 
 
 @flaky(max_runs=3)
+@pytest.mark.skipif(
+    sys.version_info[0] < 3,
+    reason="Fails with python 2")
 def test_debug_namespace(tmpdir):
     """
     Test that the kernel uses the proper namespace while debugging.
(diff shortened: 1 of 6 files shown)


spyder_kernels/console/tests/test_console_kernel.py
```diff
@@ -1143,5 +1137,46 @@ def test_get_interactive_backend(backend):
             assert value == '0'
 
 
+@flaky(max_runs=3)
+def test_debug_namespace(tmpdir):
```
ccordoba12, line 1141: It seems this test doesn't work on Python 2. For simplicity, you can skip it because I doubt many people is still using it,

ccordoba12 (commented):
Thanks for your prompt help with this @impact27! I left a small comment for you, otherwise looks good to me.

spyder_kernels/console/tests/test_console_kernel.py
impact27: I think Python2 test are generally broken. I guess pytest dropped compatibility with python 2

ccordoba12 (approved):

PR #407, opened 2022-08-03
PR: Use `get_size` to get variable length for `get_var_properties` to prevent triggering Dask tasks

dalthviz:
Addresses spyder-ide/spyder#18434


ccordoba12 (approved):

PR #412, opened 2022-09-12
PR: Use temporary file for faulthandler and make sure iopub is open for comms

impact27:
goes with https://github.com/spyder-ide/spyder/pull/19411

### DIFF
--- a/requirements/linux.yml
+++ b/requirements/linux.yml
@@ -1,5 +1,0 @@
-# Linux specific dependencies
-channels:
-  - conda-forge
-dependencies:
-  - pyxdg >=0.26
--- a/requirements/linux.yml
+++ b/requirements/linux.yml
@@ -1,0 +1,5 @@
+# Linux specific dependencies
+channels:
+  - conda-forge
+dependencies:
+  - pyxdg >=0.26
--- a/spyder_kernels/comms/frontendcomm.py
+++ b/spyder_kernels/comms/frontendcomm.py
@@ -163,8 +163,9 @@ class FrontendComm(CommBase):
             # that it will go through. Try sending it again if it passed.
             if comm_id in self._pending_comms:
                 self.notify_comm_ready(self._pending_comms[comm_id][0])
-            # remove the recieved message
+            # Remove the received message
             self.iopub_monitor.recv(flags=zmq.NOBLOCK)
+
             # There usually is a second message
             if self.iopub_monitor.poll(100):
                 self.iopub_monitor.recv(flags=zmq.NOBLOCK)
@@ -196,8 +197,9 @@ class FrontendComm(CommBase):
             msg['content']['data']['pickle_highest_protocol'])
 
         self.notify_comm_ready(comm)
+
         # The above call might fail if iopub is not connected.
-        # Check if a connection happens soon
+        # Check if a connection will happen soon
         self._wait_for_iopub_connect(comm)
 
     def _comm_close(self, msg):
--- a/spyder_kernels/comms/frontendcomm.py
+++ b/spyder_kernels/comms/frontendcomm.py
@@ -49,6 +49,7 @@ class FrontendComm(CommBase):
         self.comm_lock = threading.Lock()
         self._cached_messages = {}
         self._pending_comms = {}
+        self.iopub_monitor = None
 
     def close(self, comm_id=None):
         """Close the comm and notify the other side."""
@@ -137,12 +138,41 @@ class FrontendComm(CommBase):
                 comm.handle_msg(msg)
             self._cached_messages.pop(comm.comm_id)
 
-    def notify_iopub_ready(self):
-        """Notify frontend that the iopub is ready"""
-        for comm_id in self._pending_comms:
-            self.notify_comm_ready(self._pending_comms[comm_id])
-
     # --- Private --------
+    def _wait_for_iopub_connect(self, comm):
+        """Wait 10 seconds for iopub new connections."""
+        # This is necessary because any iopub message 
+        # (including comm responses) will be dropped before iopub connection is
+        # finished. (See https://zguide.zeromq.org/docs/chapter1 :
+        # ** the subscriber will always miss the first messages that the 
+        #    publisher sends)
+        if self.iopub_monitor is None:
+            self.iopub_monitor = self.kernel.iopub_socket.get_monitor_socket(
+                zmq.EVENT_HANDSHAKE_SUCCEEDED)
+        iopub_monitor_thread = threading.Thread(
+            target=self._iopub_monitor_main, args=(comm.comm_id, ))
+        self._pending_comms[comm.comm_id] = (comm, iopub_monitor_thread)
+        iopub_monitor_thread.start()
+
+    def _iopub_monitor_main(self, comm_id):
+        """
+        Send comm message to frontend when comms are connected
+        """
+        if self.iopub_monitor.poll(1e4):
+            # If a message is recieved on iopub_monitor, there is a good chance
+            # That it will go through. Try sending again if it passed
+            if comm_id in self._pending_comms:
+                self.notify_comm_ready(self._pending_comms[comm_id][0])
+            # remove the recieved message
+            self.iopub_monitor.recv(flags=zmq.NOBLOCK)
+            # There usually is a second message
+            if self.iopub_monitor.poll(100):
+                self.iopub_monitor.recv(flags=zmq.NOBLOCK)
+
+    def _comm_ready_callback(self, ret):
+        """A comm has replied"""
+        self._pending_comms.pop(self.calling_comm_id, None)
+
     def _wait_reply(self, comm_id, call_id, call_name, timeout, retry=True):
         """Wait until the frontend replies to a request."""
         def reply_received():
@@ -156,10 +186,6 @@ class FrontendComm(CommBase):
                 "Timeout while waiting for '{}' reply.".format(
                     call_name))
 
-    def _comm_ready_callback(self, ret):
-        """A comm has replied"""
-        self._pending_comms.pop(self.calling_comm_id, None)
-
     def _comm_open(self, comm, msg):
         """
         A new comm is open!
@@ -169,9 +195,10 @@ class FrontendComm(CommBase):
         self._set_pickle_protocol(
             msg['content']['data']['pickle_highest_protocol'])
 
-        # Frontend might not recieve comm messages, might send again later
-        self._pending_comms[comm.comm_id] = comm
         self.notify_comm_ready(comm)
+        # The above call might fail if iopub is not connected.
+        # Check if a connection happens soon
+        self._wait_for_iopub_connect(comm)
 
     def _comm_close(self, msg):
         """Close comm."""
--- a/spyder_kernels/comms/frontendcomm.py
+++ b/spyder_kernels/comms/frontendcomm.py
@@ -45,7 +45,7 @@ class FrontendComm(CommBase):
         self.kernel = kernel
         self.kernel.comm_manager.register_target(
             self._comm_name, self._comm_open)
-        self.comm_lock = threading.Lock()
+        self.comm_lock = threading.RLock()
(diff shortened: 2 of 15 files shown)


ccordoba12:
I don't know if you could rework your commits a bit to take into account PR #413.

spyder_kernels/console/kernel.py
```diff
@@ -126,30 +124,77 @@ def frontend_call(self, blocking=False, broadcast=True,
...
-    def disable_faulthandler(self):
-        """
-        Cancel the faulthandling, close the file handle and remove the file.
-        """
-        if self.faulthandler_handle:
-            faulthandler.disable()
-            self.faulthandler_handle.close()
-            self.faulthandler_handle = None
+        self.faulthandler_handle = tempfile.NamedTemporaryFile(
+            'wt', suffix='.fault')
```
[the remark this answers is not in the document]
impact27: I am not sure how to do that? On mac at least this create a file like: `'/var/folders/xx/mbs7xxcs2397g51c4kbg23p00000gp/T/tmps5mzxk5n.fault'`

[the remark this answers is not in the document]
impact27: It doesn't look like this can be easily done with `tempfile.NamedTemporaryFile`. We could add linux specific code but that would add complexity to the code and possible bugs down the line.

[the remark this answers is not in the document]
impact27: Maybe I should expand:
 - What I need here is a temporary file that would be destroyed as soon as the kernel stops. The only way for the file to not disappear should be for the kernel to crash.
 - `tempfile.NamedTemporaryFile` Fits this description perfectly and should work as advertised for all OS, but doesn't allow the automatic creation of folders.
 - `tempfile.TemporaryDirectory` would not work because it would create a directory for each kernel.
 - The solution would be to create for linux `tmp/spyder-kernels`but that means creating a directory that might not be deletable easily. Also this solution would be os-specific.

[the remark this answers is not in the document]
ccordoba12: > The solution would be to create for linux /tmp/spyder-kernelsbut that means creating a directory that might not be deletable easily. Also this solution would be os-specific.

I understand, but I'm really against polluting `/tmp` on Linux (as many apps do these days). So please implement a specific solution for that OS.

To do that, I suggest to add as a new dependency the `pyxdg` package, use

https://pyxdg.readthedocs.io/en/latest/basedirectory.html#xdg.BaseDirectory.xdg_data_home

to get the data directory per user, create or check that `spyder` exists under it and use it to save these files. Only if that fails, you should resort to your current approach of saving them directly in `/tmp`.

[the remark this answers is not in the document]
impact27: Alternatively I could create `~/.local/share` or `/tmp/spyder` and save the files in this folder?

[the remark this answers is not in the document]
ccordoba12: > Alternatively I could create ~/.local/share

That's the default XDG directory but it can be changed by setting an env var.

> /tmp/spyder and save the files in this folder?

That wouldn't work on a multi-user system because different users are not able to write to the same directory.

So, I think it's simpler and better to use `pyxdg`, which is a very lightweight dependency, and avoid to handle that ourselves.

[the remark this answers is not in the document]
impact27: Done but I am unsure if the dependency is correctly added for conda

spyder_kernels/comms/frontendcomm.py
```diff
@@ -127,22 +127,52 @@ def cache_message(self, comm_id, msg):
...
-    def notify_iopub_ready(self):
-        """Notify frontend that the iopub is ready"""
-        self._iopub_ready = True
-        for comm in self._pending_comms:
-            self.notify_comm_ready(comm)
-        self._pending_comms = []
-
     # --- Private --------
+    def _wait_for_iopub_connect(self, comm):
+        """Wait 10 seconds for iopub new connections."""
```
ccordoba12, line 143: I'm not sure if this docstring is correct because I don't clearly see the 10 seconds in this method's code. But even if this is correct, perhaps we should wait longer on Windows because the kernel usually takes more time to start there (perhaps 20 or 30 secs?).

spyder_kernels/comms/frontendcomm.py
```diff
@@ -127,22 +127,52 @@ def cache_message(self, comm_id, msg):
...
+        self._pending_comms[comm.comm_id] = (comm, iopub_monitor_thread)
+        iopub_monitor_thread.start()
+
+    def _iopub_monitor_main(self, comm_id):
+        """
+        Send comm message to frontend when comms are connected
+        """
+        if self.iopub_monitor.poll(1e4):
+            # If a message is recieved on iopub_monitor, there is a good chance
+            # That it will go through. Try sending again if it passed
```
ccordoba12, line 163:
```suggestion
            # If a message is received on iopub_monitor, there is a good chance
            # that it will go through. Try sending it again if it passed.
```

spyder_kernels/comms/frontendcomm.py
```diff
@@ -127,22 +127,52 @@ def cache_message(self, comm_id, msg):
...
+            if comm_id in self._pending_comms:
+                self.notify_comm_ready(self._pending_comms[comm_id][0])
+            # remove the recieved message
+            self.iopub_monitor.recv(flags=zmq.NOBLOCK)
+            # There usually is a second message
```
ccordoba12, line 168:
```suggestion
            # Remove the received message
            self.iopub_monitor.recv(flags=zmq.NOBLOCK)

            # There usually is a second message
```

spyder_kernels/comms/frontendcomm.py
```diff
@@ -165,11 +195,10 @@ def _comm_open(self, comm, msg):
         self._set_pickle_protocol(
             msg['content']['data']['pickle_highest_protocol'])
 
-        if not self._iopub_ready:
-            # Frontend will not recieve comm messages
-            self._pending_comms.append(comm)
-            return
         self.notify_comm_ready(comm)
+        # The above call might fail if iopub is not connected.
+        # Check if a connection happens soon
```
ccordoba12, line 200:
```suggestion

        # The above call might fail if iopub is not connected.
        # Check if a connection will happen soon
```

requirements/linux.yml
```diff
@@ -0,0 +1,5 @@
+# Linux specific dependencies
+channels:
+  - conda-forge
+dependencies:
+  - pyxdg >=0.26
```
ccordoba12, line 5: Please remove this file and instead add the last line to  `requirements/posix.txt`.

spyder_kernels/console/kernel.py
```diff
@@ -126,30 +124,90 @@ def frontend_call(self, blocking=False, broadcast=True,
...
-            self.faulthandler_handle = None
+        fault_dir = None
+        if sys.platform.startswith('linux'):
+            # Do not use /tmp for temporary files
+            try:
+                from xdg.BaseDirectory import xdg_data_home
+                fault_dir = xdg_data_home
+                os.makedirs(fault_dir, exist_ok=True)
+            except Exception:
+                fault_dir = None
```
ccordoba12, line 140:
```suggestion
                fault_dir = '/tmp'
```

I think this could happen in very minimalist Linux environments like Docker containers.

spyder_kernels/console/shell.py
```diff
@@ -64,10 +64,10 @@ def __init__(self, *args, **kwargs):
 
     def ask_exit(self):
         """Engage the exit actions."""
-        if self.active_eventloop != "inline":
+        if self.active_eventloop not in [None, "inline"]:
             # Some eventloops prevent the kernel from shutting down
             self.enable_gui('inline')
-        return super(SpyderShell, self).ask_exit()
+        super(SpyderShell, self).ask_exit()
```
ccordoba12, line 70:
```suggestion
        super().ask_exit()
```

Use `super` in the way allowed by Python 3.

spyder_kernels/console/kernel.py
```diff
@@ -27,7 +29,7 @@
 from zmq.utils.garbage import gc
 
 # Local imports
-from spyder_kernels.comms.frontendcomm import FrontendComm
+from spyder_kernels.comms.frontendcomm import FrontendComm, CommError
```
ccordoba12, line 32:
```suggestion
from spyder_kernels.comms.frontendcomm import FrontendComm
```

Seems to not be used anymore.

spyder_kernels/console/kernel.py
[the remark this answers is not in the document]
ccordoba12: Thanks, I'll leave a comment so you add it to the right file.

ccordoba12 (commented):
@impact27, I'm not sure if this ready for review yet, but I'm leaving a few comments/suggestions for you below.

ccordoba12:
@impact27, please update the PR title to describe the changes you did in your last commits because they are not related to the faulthandler file.

impact27:
@ccordoba12 many thanks for the comments. I noticed that sometimes the fault text was not printed. While debugging the issue I found out that the iopub channel drops io and comm messages during startup (even though the shell channel is up) so I am trying to fix that. The current state decreases the probability of dropping messages but I still see some.

spyder_kernels/comms/frontendcomm.py
impact27: The 10 seconds are in the ` _iopub_monitor_main`
The problem here is that I have two options:
 - I can wait indefinitely, but then the shutdown will be problematic
 - I can have a timeout, but then I need to find a way of handling the case where no messages are received.

spyder_kernels/console/kernel.py
impact27: None means the default directory is used

requirements/linux.yml
impact27: Wouldn't that affect macos as well?

impact27:
@ccordoba12 This should be ready now. I dropped the socket monitor approach as it was not reliable. The kernel comm will now send messages to the frontend every second until the frontend replies.

impact27:
@ccordoba12 should I separate the PR in two or are you happy to review it together?

spyder_kernels/comms/frontendcomm.py
```diff
@@ -124,6 +124,35 @@ def cache_message(self, comm_id, msg):
...
+
+    def _notify_comm_ready(self, comm):
+        """Send messages about comm readiness to frontend."""
+        self.remote_call(
+            comm_id=comm.comm_id,
+            callback=self._comm_ready_callback
+        )._comm_ready()
+
+    def _comm_ready_callback(self, ret):
+        """A comm has replied"""
```
ccordoba12, line 145:
```suggestion
        """A comm has replied, so process all cached messages related to it."""
```

I think that's what this is doing, right?

spyder_kernels/comms/frontendcomm.py
```diff
@@ -145,13 +174,12 @@ def _comm_open(self, comm, msg):
...
             msg['content']['data']['pickle_highest_protocol'])
-        self.remote_call()._set_pickle_protocol(pickle.HIGHEST_PROTOCOL)
-        # Handle cached messages
-        if comm.comm_id in self._cached_messages:
-            for msg in self._cached_messages[comm.comm_id]:
-                comm.handle_msg(msg)
-            self._cached_messages.pop(comm.comm_id)
 
+        # IOPub might not be connected yet, keep sending messages until a
+        # reply is recieved
```
ccordoba12, line 179:
```suggestion
        # reply is received.
```

ccordoba12:
> @ccordoba12 should I separate the PR in two or are you happy to review it together?

I'm fine with having both changes in the same PR.

ccordoba12:
@impact27, I'll merge this after @dalthviz releases 5.4.0 to not break the release workflow.

ccordoba12 (approved):
Thanks @impact27!

PR #413, opened 2022-09-19
PR: Flush std streams after execution

impact27:
5.x version of https://github.com/spyder-ide/spyder-kernels/pull/412


ccordoba12 (approved):

PR #415, opened 2022-09-22
PR: Fix tk eventloop handling when debugging on Windows

dalthviz:
Part of spyder-ide/spyder#17523

### DIFF
--- a/spyder_kernels/customize/spyderpdb.py
+++ b/spyder_kernels/customize/spyderpdb.py
@@ -10,6 +10,7 @@
 import ast
 import bdb
 import logging
+import os
 import sys
 import traceback
 import threading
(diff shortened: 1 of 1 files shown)


spyder_kernels/customize/spyderpdb.py
```diff
@@ -703,7 +704,13 @@ def cmd_input(self, prompt=''):
         if is_main_thread and kernel.eventloop:
             while self._cmd_input_line is None:
                 eventloop = kernel.eventloop
-                if eventloop:
+                # Check if the current backend is Tk on Windows
+                # to let GUI update.
+                # See spyder-ide/spyder#17523
+                if (eventloop and hasattr(kernel, "app_wrapper") and
+                        os.name == "nt"):
```
ccordoba12, line 711: Did you check that this is not necessary for Windows or Mac?

ccordoba12 (commented):
Thanks for the finding the solution to this nasty bug @dalthviz! Other my simple question below, this looks good to me.

spyder_kernels/customize/spyderpdb.py
dalthviz: Checked on Windows and Linux. On Windows the change is necessary on Linux no. I think the reason is because on Linux/Mac the tk eventloop is used with the `createfilehandler` method which is not available for Windows. Should I test on Mac too?

ccordoba12: Yeah, please do. I meant Linux and Mac above (sorry for the confusion).

dalthviz: Checked and seems like is not necessary with Mac (so the change is only necessary for Windows) @ccordoba12

ccordoba12: Ok, great! Merging then.

ccordoba12 (approved):
Thanks @dalthviz!

PR #426, opened 2022-10-27
PR: Update ipykernel constraint to >=6.16.1

dalthviz:
Part of spyder-ide/spyder#19874


ccordoba12:
@dalthviz, I'm going to push a commit to your branch to fix the failing test. The failure is caused by PR https://github.com/ipython/ipykernel/pull/1008, which is precisely the change we want to depend on by increasing IPykernel to `6.16.1` since it fixes a quite serious memory leak.

PR #424, opened 2022-10-19
PR: Patch for CVE-2007-4559 Tar directory traversal

TrellixVulnTeam:
# Patching CVE-2007-4559

Hi, we are security researchers from the Advanced Research Center at [Trellix](https://www.trellix.com). We have began a campaign to patch a widespread bug named CVE-2007-4559. CVE-2007-4559 is a 15 year old bug in the Python tarfile package. By using extract() or extractall() on a tarfile object without sanitizing input, a maliciously crafted .tar file could perform a directory path traversal attack. We found at least one unsantized extractall() in your codebase and are providing a patch for you via pull request. The patch essentially checks to see if all tarfile members will be extracted safely and throws an exception otherwise. We encourage you to use this patch or your own solution to secure against CVE-2007-4559. Further technical information about the vulnerability can be found in this [blog](https://www.trellix.com/en-us/about/newsroom/stories/research/tarfile-exploiting-the-world.html).

If you have further questions you may contact us through this projects lead researcher [Kasimir Schulz](mailto:kasimir.schulz@trellix.com).

PR #428, opened 2022-11-02
PR: Patch for CVE-2007-4559 Tar directory traversal

ccordoba12:
From the original pull request (i.e. PR #424):

## Patching CVE-2007-4559

Hi, we are security researchers from the Advanced Research Center at [Trellix](https://www.trellix.com). We have began a campaign to patch a widespread bug named CVE-2007-4559. CVE-2007-4559 is a 15 year old bug in the Python tarfile package. By using extract() or extractall() on a tarfile object without sanitizing input, a maliciously crafted .tar file could perform a directory path traversal attack. We found at least one unsantized extractall() in your codebase and are providing a patch for you via pull request. The patch essentially checks to see if all tarfile members will be extracted safely and throws an exception otherwise. We encourage you to use this patch or your own solution to secure against CVE-2007-4559. Further technical information about the vulnerability can be found in this [blog](https://www.trellix.com/en-us/about/newsroom/stories/research/tarfile-exploiting-the-world.html).

If you have further questions you may contact us through this projects lead researcher [Kasimir Schulz](mailto:kasimir.schulz@trellix.com).

### DIFF
--- a/spyder_kernels/utils/iofuncs.py
+++ b/spyder_kernels/utils/iofuncs.py
@@ -23,7 +23,6 @@ import tarfile
 import tempfile
 import shutil
 import types
-import warnings
 import json
 import inspect
 import dis
@@ -370,6 +369,27 @@ def save_dictionary(data, filename):
     return error_message
 
 
+def is_within_directory(directory, target):
+    """Check if a file is within a directory."""
+    abs_directory = os.path.abspath(directory)
+    abs_target = os.path.abspath(target)
+    prefix = os.path.commonprefix([abs_directory, abs_target])
+    return prefix == abs_directory
+
+
+def safe_extract(tar, path=".", members=None, numeric_owner=False):
+    """Safely extract a tar file."""
+    for member in tar.getmembers():
+        member_path = os.path.join(path, member.name)
+        if not is_within_directory(path, member_path):
+            raise Exception(
+                "Attempted path traversal in tar file {}".format(
+                    repr(tar.name)
+                )
+            )
+    tar.extractall(path, members, numeric_owner=numeric_owner)
+
+
 def load_dictionary(filename):
     """Load dictionary from .spydata file"""
     filename = osp.abspath(filename)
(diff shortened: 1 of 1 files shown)


spyder_kernels/utils/iofuncs.py
```diff
@@ -370,6 +369,25 @@ def save_dictionary(data, filename):
...
+    return prefix == abs_directory
+
+
+def safe_extract(tar, path=".", members=None, numeric_owner=False):
+    """Safely extract a tar file."""
+    for member in tar.getmembers():
+        member_path = os.path.join(path, member.name)
+        if not is_within_directory(path, member_path):
+            raise Exception(
+                r"Attempted path traversal in tar file {}".format(tar.name)
```
CAM-Gerlach, line 386:
```suggestion
                f"Attempted path traversal in tar file {tar.name!r}"
```

* Not sure why this is a raw string, as no escapes are used
* May as well use an f-string instead of `format`?
* Using the `r` (repr) format code will quote the file name

CAM-Gerlach (commented):
One comment, otherwise LGTM (as far as I'm qualified to review this)

spyder_kernels/utils/iofuncs.py
ccordoba12: > Not sure why this is a raw string, as no escapes are used

Sorry, I thought the `!r` meant raw string.

> May as well use an f-string instead of format?

The 2.x branch still support Python 2, so that's not possible.

> Using the r (repr) format code will quote the file name

Good to know! That's a change I can make.

CAM-Gerlach: > Sorry, I thought the !r you suggested in other PR meant raw string.

`!` indicates format codes in the format-string mini language, analogous to `%f`, `%s`, etc. in printf/`%`-style formatting

> The 2.x branch still support Python 2, so that's not possible.

Oh, didn't think of that even though the code had a `PY2` branch; its been a while since I've written Python 2 compatible code, and even if the function is never called on Py2, the unrecognized string prefix will presumably raise a SyntaxError in the tokenizer or the parser before it is compiled to bytecode and the AST won't be valid. Good catch.

> Good to know! That's a change I can make.

What I meant was using the proper format code as I did above, i.e.:

```python
raise Exception("Attempted path traversal in tar file {!r}".format(tar.name))
```

Like the other `format` syntax, it works the same whether you're using f-strings or `str.format()`.

Issue #425, opened 2022-10-20
Possible minor issues related to post mortem debugging


rear1019:
We have noticed two minor issues related to post mortem debugging¹, though this might be intended:
* When an error occurs, the post mortem debugger is invoked and the current stack frame is printed in the IPython Console. However, the editor does _not_ jump to the position of the error. Entering any command in the IPython Console makes the editor jump to the position of the error. Fix available.
* Post mortem debugging is not entered if “Debug file” instead of “Run file” is used. We do have a simple fix, but the fix only works if no recursive debugging is involved.

¹ option “Directly enter debugging when errors appear” in Spyder’s run configuration dialog

EDIT: Spyder v5.3.3 and spyder-kernels v2.3.3. Issue created here as these are bugs in spyder-kernels, not Spyder.

ccordoba12:
Hey @rear1019, thanks for reporting. You said

> When an error occurs, the post mortem debugger is invoked and the current stack frame is printed in the IPython Console. However, the editor does not jump to the position of the error. Entering any command in the IPython Console makes the editor jump to the position of the error. Fix available.

Great! Please submit a PR for it.

> Post mortem debugging is not entered if “Debug file” instead of “Run file” is used. We do have a simple fix, but the fix only works if no recursive debugging is involved.

Ok, I'm not so sure about that but we could see if it's applicable after looking at the PR.

rear1019:
PR for first issue: #430 

Regarding the second issues: Make use of the `post_mortem` argument passed to `debugfile()` (contained in `spydercustomize.py`). It works if _no_ recursive debugging is involved (`_exec_file()` call in `debugfile()`), but fails with recursive debugging depending on the interaction with the debugger.

ccordoba12:
@impact27, could you take a look at this issue? Thanks!

PR #430, opened 2022-11-14
PR: Inform GUI about position of exception in post mortem debugging

rear1019:
See commit message for details. Fixes one of the issues mentioned in #425.


ccordoba12:
This looks good to me for `2.x`, but I see no equivalent of `pdb.send_initial_notification` in `master`.

@impact27, how this work should be ported to master?

impact27:
Is this bug still present in master? I can't reproduce in master

ccordoba12:
You're right, I just tested and this bug doesn't show up in master.

PR #434, opened 2022-12-28
PR: Use `allow_pickle=True` when loading Numpy arrays

nkleinbaer:
Pass `allow_pickle=True` when calling `np.load` to enable loading spydata files that contain numpy arrays of type object (see https://github.com/spyder-ide/spyder/issues/11754)


ccordoba12 (commented):
Hey @nkleinbaer, thanks a lot for your contribution! This looks good to me, but I don't understand why you added the `export_data.spydata` and  `export_data_renamed.spydata` files, given that you don't seem to use them in our tests.

If they are not really necessary, please remove them and rebase this PR on top of the latest `2.x` to not have conflicts in your Spyder PR.

nkleinbaer:
Github makes it look like I added those as new files, but they actually already exist and are used here:
https://github.com/spyder-ide/spyder-kernels/blob/e34e190bc40fe7db1bca7c44a7a024d0c6d1f346/spyder_kernels/utils/tests/test_iofuncs.py#L230-L244

The reason I replaced those files is because I changed the `spydata_values` fixture to include an object array, so I needed to update the test data to match that.

nkleinbaer:
Also, the updated `export_data.spydata` file was created by running the following code in Spyder and using the save button in the variable explorer. 

```python
A = 1
B = 'ham'
C = np.eye(3)
D = {'a': True, 'b': np.eye(4, dtype=np.complex128)}
E = [np.eye(2, dtype=np.int64), 42.0, np.eye(3, dtype=np.bool_), np.eye(4, dtype=object)]
```

As I think about it now, though, I don't know of anyway for you to independently verify that is what the file actually contains. And given my main point in https://github.com/spyder-ide/spyder/pull/20272 is that `spydata` files can potentially contain malicious code, it would be understandable if you are skeptical of these files :laughing: . Let me know if that is a concern.

CONDUCTOR77747:
Amazing! Thank you! (I understand the risks. I only need to get around this problem once. Then I have to put everything back the way it was.)

ccordoba12:
Closing and reopening to restart our tests (which didn't run in the last commit).

ccordoba12:
Ok, some tests are failing on Python 2 (yes, we still support it in this package). So, I'm to skip them in order to merge this PR.

ccordoba12 (approved):
Thanks @nkleinbaer for your contribution!

PR #435, opened 2023-01-06
PR: Replace deprecated `inspect.formatargspec` call

### LINKED ISSUE #307: formatargspec is deprecated since Python 3.5
While using spyder 4.2.5 from conda-forge, I get the message:
> `dochelpers.py`:125: DeprecationWarning: `formatargspec` is deprecated since Python 3.5. Use `signature` and the `Signature` object directly

The documentation for [`inspect.formatargspec`](https://docs.python.org/3/library/inspect.html#inspect.formatargspec) reflects this warning.

The relevant code to potentially refactor is: https://github.com/spyder-ide/spyder-kernels/blob/7cbd99673a03dc732041c1d3fad7795c1b683b58/spyder_kernels/utils/dochelpers.py#L116-L130

If Python 3.5 is no longer supported here, then hopefully it'll be straight-forward to do. Otherwise, a third block would be need to handle Python <3.5 and >=3.5 versions.

juliangilbey:
I have based this patch on the 2.x branch rather than the master branch; the patch would need the obvious modification to apply there (dedenting and changing the surrounding context).

Fixes #307

### DIFF
--- a/spyder_kernels/utils/dochelpers.py
+++ b/spyder_kernels/utils/dochelpers.py
@@ -120,11 +120,8 @@ def getdoc(obj):
                     args, varargs, varkw, defaults,
                     formatvalue=lambda o:'='+repr(o))
             else:
-                (args, varargs, varkw, defaults, kwonlyargs, kwonlydefaults,
-                 annotations) = inspect.getfullargspec(obj)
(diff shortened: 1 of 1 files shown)


ccordoba12 (approved):
Looks good to me, thanks @juliangilbey!

PR #436, opened 2023-01-07
replacing formatargspec() per inspect.signature()

stonebig:
The formatargspec() function is deprecated since Python 3.5, and removed in Python-3.11
It would be nice to have a fix in coming Spyder-5.4.2

## Description of Changes

* I replace The formatargspec() function, deprecated since Python 3.5; per [inspect.signature()](https://docs.python.org/3/library/inspect.html#inspect.signature)
* I checked on lambda and not lambda 
image

image

### DIFF
--- a/spyder_kernels/utils/dochelpers.py
+++ b/spyder_kernels/utils/dochelpers.py
@@ -110,11 +110,7 @@ def getdoc(obj):
             doc['note'] = 'Function'
         doc['name'] = obj.__name__
         if inspect.isfunction(obj):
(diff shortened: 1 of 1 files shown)


stonebig:
was originaly posted on [spyder_ide/spyder](https://github.com/spyder-ide/spyder/pull/20322), but I guess it's not the right place

ccordoba12:
Thanks @stonebig for your help with this! However, I'm afraid I'll have to close it as a duplicate of PR #435, which basically provides the same solution and came first.

PR #437, opened 2023-01-12
PR: Remove imports from __future__

oscargus:
It seems like PY2 is still in 2.x, but not master, so I guess that master is the correct branch to target for this?

Just a minor cleanup.


ccordoba12:
> It seems like PY2 is still in 2.x, but not master, so I guess that master is the correct branch to target for this?

That's correct.

ccordoba12 (approved):
Thanks for your help with this @oscargus!

PR #439, opened 2023-01-20
Add FreeBSD to test user site-packages in pathlist

rhurlin:
This is the only one of the test suite that FreeBSD does not pass, it was not recognized. Since FreeBSD uses the same path as Darwin, it can simply be appended to the if statement.

### DIFF
--- a/spyder_kernels/customize/tests/test_utils.py
+++ b/spyder_kernels/customize/tests/test_utils.py
@@ -16,7 +16,7 @@ def test_user_sitepackages_in_pathlist():
     """Test that we include user site-packages in pathlist."""
     if sys.platform.startswith('linux'):
         user_path = 'local'
-    elif (sys.platform.startswith('darwin') | sys.platform.startswith('freebsd')):
+    elif (sys.platform = 'darwin' or sys.platform.startswith('freebsd')):
         user_path = os.path.expanduser('~/.local')
     else:
         user_path = 'Roaming'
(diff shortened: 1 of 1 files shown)


spyder_kernels/customize/tests/test_utils.py
```diff
@@ -16,7 +16,7 @@ def test_user_sitepackages_in_pathlist():
     """Test that we include user site-packages in pathlist."""
     if sys.platform.startswith('linux'):
         user_path = 'local'
-    elif sys.platform == 'darwin':
+    elif (sys.platform.startswith('darwin') | sys.platform.startswith('freebsd')):
```
ccordoba12, line 19: I suggest to change this to

```python
elif (sys.platform == darwin or sys.platform.startswith('freebsd')):
```

This conforms to the way we check for Mac in our codebase.

ccordoba12 (commented):
Hey @rhurlin, thanks for your contribution! I left a small suggestion for you, otherwise looks good to me.

Also, since you made this PR against our `master` branch, it would be part of Spyder-kernels 3, to be released along with Spyder 6 later this year. If you'd like to see it in our next minor version (2.4.3) instead, please redo it against our `2.x` branch.

spyder_kernels/customize/tests/test_utils.py
rhurlin: Thanks for your fast and helpful response. Of course, I will follow you suggestion about darwin.

If I want to submit a PR against the 2.x branch, should I open a new PR or is it possible to do it in this one?

ccordoba12: I'm afraid you have to open a new PR against `2.x` and close this one. That's because you didn't use a branch to create your PR but instead pushed directly to `master` (with a branch it'd be a matter of issuing a few git commands to make it derive from `2.x`)

rhurlin:
Ok, thanks for clarification. I close this PR and will open an new one in the 2.x branch.

PR #441, opened 2023-01-22
Add FreeBSD to test user site-packages in pathlist

rhurlin:
The 'user site-packages in pathlist' test is the only one that fails on FreeBSD because the OS is not recognized.  FreeBSD uses the same .local path as Darwin, so you can simply extend the if statement.

This PR is a continuation of #439, which I closed because it was directed against the master branch instead of 2.x.


rhurlin:
Obviously this commit went wrong. The plan was to commit a single change from my fork (master) to the 2.x branch of this repo.
So I close this PR and try again, sorry!

ccordoba12:
@rhurlin, you need to do the following:

```
git checkout 2.x
git checkout -b freebsd-fixes
git commit <your changes>
git push origin freebsd-fixes
```

rhurlin:
> ```
> git checkout 2.x
> git checkout -b freebsd-fixes
> git commit <your changes>
> git push origin freebsd-fixes
> ```

@ccordoba12, Thanks for the guidance. I still struggle with GitHub, having worked with subversion for many years.

I already tried to check out a branch 2.x on my fork, but got the answer that there is no such branch (in `rhurlin/spyder-kernels`). But directly in the original repo (`spyder-ide/spyder-kernels`) I am not allowed to commit. So somehow I'm on the hose and get no further :(

PR #442, opened 2023-01-24
PR: Add FreeBSD to `test_user_sitepackages_in_pathlist`

rhurlin:
The 'user site-packages in pathlist' test is the only one that fails on FreeBSD because the OS is not recognized. FreeBSD uses the same .local path as Darwin, so you can simply extend the if statement.

### DIFF
--- a/spyder_kernels/customize/tests/test_utils.py
+++ b/spyder_kernels/customize/tests/test_utils.py
@@ -16,7 +16,7 @@ def test_user_sitepackages_in_pathlist():
     """Test that we include user site-packages in pathlist."""
     if sys.platform.startswith('linux'):
         user_path = 'local'
-    elif sys.platform == 'darwin':
+    elif (sys.platform == 'darwin' or sys.platform.startswith('freebsd'):
         user_path = os.path.expanduser('~/.local')
     else:
         user_path = 'Roaming'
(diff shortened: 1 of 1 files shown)


spyder_kernels/customize/tests/test_utils.py
```diff
@@ -16,7 +16,7 @@ def test_user_sitepackages_in_pathlist():
     """Test that we include user site-packages in pathlist."""
     if sys.platform.startswith('linux'):
         user_path = 'local'
-    elif sys.platform == 'darwin':
+    elif (sys.platform == 'darwin' or sys.platform.startswith('freebsd'):
```
ccordoba12, line 19:
```suggestion
    elif (sys.platform == 'darwin' or sys.platform.startswith('freebsd')):
```

The initial parenthesis was not closed at the end of the statement.

ccordoba12 (commented):
Thanks for your contribution @rhurlin! I left a small suggestion for you below, otherwise looks good to me.

In case you don't know how to work with suggestions on Github, please read [these docs](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/incorporating-feedback-in-your-pull-request).

rhurlin:
Sorry, this should not have happened.

ccordoba12 (approved):
Thanks @rhurlin!

PR #452, opened 2023-03-31
PR: Fix error when executing empty Python script

rear1019:


ccordoba12 (approved):
Tested locally and it works as expected. Thanks @rear1019!

PR #460, opened 2023-06-13
PR: Add a global filter flag to settings

jsbautista:
Part of https://github.com/spyder-ide/spyder/pull/20868

### DIFF
--- a/spyder_kernels/console/tests/test_console_kernel.py
+++ b/spyder_kernels/console/tests/test_console_kernel.py
@@ -305,11 +305,6 @@ def test_get_namespace_view_filter_on(kernel, filter_on):
     if not filter_on:
         assert 'TestFilterOn' in nsview
         assert 'TestFilterOff' in nsview
-        assert "'type': 'int'" in nsview or "'type': u'int'" in nsview
-        assert "'size': 1" in nsview
-        assert "'view': '1'" in nsview
-        assert "'numpy_type': 'Unknown'" in nsview
-        assert "'python_type': 'int'" in nsview
     else:
         assert 'TestFilterOff' not in nsview
         assert 'TestFilterOn' not in nsview
--- a/spyder_kernels/console/tests/test_console_kernel.py
+++ b/spyder_kernels/console/tests/test_console_kernel.py
@@ -294,7 +294,7 @@ def test_get_namespace_view_filter_on(kernel, filter_on):
     """
     Test the namespace view of the kernel with filters on and off.
     """
-    execute = asyncio.run(kernel.do_execute('TestFilterOn = 1', True))
+    execute = asyncio.run(kernel.do_execute('a = 1', True))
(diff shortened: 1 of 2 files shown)


ccordoba12:
@jsbautista or @dalthviz, could you add a test for this? It could be as simple as building the namespace view with and without the filter on and checking that a specific variable is absent/present in each case.

spyder_kernels/utils/nsview.py
```diff
@@ -589,7 +589,8 @@ def is_callable_or_module(value):
 def globalsfilter(input_dict, check_all=False, filters=None,
                   exclude_private=None, exclude_capitalized=None,
                   exclude_uppercase=None, exclude_unsupported=None,
-                  excluded_names=None, exclude_callables_and_modules=None):
+                  excluded_names=None, exclude_callables_and_modules=None,
+                  filter_on=False):
```
dalthviz, line 593: Lets make this new kwarg follow the default value:

```suggestion
                  filter_on=True):
```

dalthviz, line 593: Lets make this new kwarg follow the default value:

```suggestion
                  filter_on=True):
```

dalthviz (commented):
Thanks @jsbautista for the work here! Left a smal suggestion. Regarding the test, I think we can use as boilerplate the one at:

https://github.com/spyder-ide/spyder-kernels/blob/fd7a942a49376aa05006907314dab37b93f66c2f/spyder_kernels/console/tests/test_console_kernel.py#L277-L289


If you have any question let me know!

spyder_kernels/console/tests/test_console_kernel.py
```diff
@@ -288,6 +289,31 @@ def test_get_namespace_view(kernel):
     assert "'python_type': 'int'" in nsview
 
 
+@pytest.mark.parametrize("filter_on", [True, False])
+def test_get_namespace_view_filter_on(kernel, filter_on):
+    """
+    Test the namespace view of the kernel with filters on and off.
+    """
+    execute = asyncio.run(kernel.do_execute('TestFilterOn = 1', True))
+    asyncio.run(kernel.do_execute('TestFilterOff = 1', True))
```
dalthviz, line 298: Not totally sure but we just need one of this variables, right? Other idea could be to define one that variable that always is shown regardelss of the filters (executing something like `a = 1`) and another one that is only shown when the filters are off (it could be `TestFilterOff`). What do you think @jsbautista ?

jsbautista: > 

You are right, I agree, I will make the necessary changes

dalthviz (approved):
Thanks @jsbautista ! LGTM :+1:

ccordoba12:
@dalthviz, please squash-merge as soon as @jsbautista resolves your comment above.

PR #461, opened 2023-06-13
PR: Increase minimal required version of IPykernel to 6.23.2

ccordoba12:
This is necessary to address spyder-ide/spyder#21017.


ccordoba12:
This requires https://github.com/conda-forge/ipykernel-feedstock/pull/163 first.

PR #464, opened 2023-07-23
Remove locals inspection

impact27:
This is necessary to address https://github.com/spyder-ide/spyder/issues/21149

goes with https://github.com/spyder-ide/spyder/pull/21184


ccordoba12 (approved):
Thanks @impact27!

PR #465, opened 2023-07-27
Save temporary file in test to temporary location

juliangilbey:
The test `test_save_load_hdf5_files` saves its test file `test.h5` in the current directory, which is not helpful.  (If you run the tests prior to installing the package, the h5 file may end up in the resulting package, for example.)

This patch saves the `test.h5` in a temporary directory instead.


ccordoba12 (approved):
Thanks @juliangilbey!

PR #467, opened 2023-09-11
PR: Fix index when skipping hidden frames (Debugger)

impact27:
Goes with https://github.com/spyder-ide/spyder/pull/21323


ccordoba12 (approved):
Thanks @impact27!

PR #476, opened 2023-11-11
PR: Send back pickling error correctly

impact27:
if an error occurred while unpickling this will send it back correctly


ccordoba12 (approved):
Thanks @impact27!

PR #477, opened 2023-11-29
PR: Handle new Inline backend options `fontsize` and `bottom`

jitseniesen:
Handle the new options for the font size and the bottom edge for inline plots, which are added in PR spyder-ide/spyder#21566.


ccordoba12 (approved):
Looks good to me, thanks @jitseniesen!

PR #487, opened 2024-05-16
PR: Fix issue where Spyder's inline graphics preferences were not applied

mrclary:
This fixes an issue where:
* Spyder's inline graphics preferences were not applied when toggling from inline to interactive and back to inline.
* Matplotlib's `rcParams` are over-written by Spyder's inline preferences even with the preference set for an interactive backend.

For more detail, see spyder-ide/spyder#22034.

### DIFF
--- a/spyder_kernels/console/kernel.py
+++ b/spyder_kernels/console/kernel.py
@@ -25,7 +25,7 @@ import cloudpickle
 # Third-party imports
 from ipykernel.ipkernel import IPythonKernel
 from ipykernel import get_connection_info
-from traitlets.config.loader import LazyConfigValue
+from traitlets.config.loader import Config, LazyConfigValue
 import zmq
 from zmq.utils.garbage import gc
 
@@ -554,13 +554,6 @@ class SpyderKernel(IPythonKernel):
         fontsize_n = 'pylab/inline/fontsize'
         bottom_n = 'pylab/inline/bottom'
         bbox_inches_n = 'pylab/inline/bbox_inches'
-        inline_backend = 'inline'
-
-        if pylab_autoload_n in conf or pylab_backend_n in conf:
-            self._set_mpl_backend(
-                conf.get(pylab_backend_n, inline_backend),
-                pylab=conf.get(pylab_autoload_n, False)
-            )
 
         if figure_format_n in conf:
             self._set_config_option(
@@ -589,6 +582,30 @@ class SpyderKernel(IPythonKernel):
         if bbox_inches_n in conf:
             self.set_mpl_inline_bbox_inches(conf[bbox_inches_n])
 
+        # To update rcParams in inline mode we can either do so directly or
+        # re-assert inline mode. However, either will prevent restoring
+        # rcParams when toggling to interactive mode. The workaround is to
+        # first toggle to interactive mode, then back to inline. This updates
+        # the rcParams and restores rcParams when switching to interactive.
+        interactive_backend = self.get_mpl_interactive_backend()
+        current_backend = self.get_matplotlib_backend()
+        pylab_changed = pylab_autoload_n in conf
+        backend_changed = pylab_backend_n in conf
+        if (
+            current_backend == 'inline'
+            and not backend_changed
+            and interactive_backend not in ('inline', -1)
+        ):
+            self._set_mpl_backend(interactive_backend)
+        if (
+            current_backend == 'inline'  # toggle back to inline for rcParams
+            or pylab_changed
+            or backend_changed
+        ):
+            self._set_mpl_backend(
+                conf.get(pylab_backend_n, current_backend),
+                pylab=conf.get(pylab_autoload_n, False)
+            )
 
     def set_mpl_inline_bbox_inches(self, bbox_inches):
         """
@@ -974,14 +991,26 @@ class SpyderKernel(IPythonKernel):
 
     def _set_mpl_inline_rc_config(self, option, value):
         """
-        Update any of the Matplolib rcParams given an option and value.
+        Update InlineBackend.rc given an option and value.
         """
-        try:
-            from matplotlib import rcParams
-            rcParams[option] = value
-        except Exception:
-            # Needed in case matplolib isn't installed
-            pass
+        _rc = {option: value}
+        if (
+            'InlineBackend' in self.config
+            and 'rc' in self.config['InlineBackend']
+        ):
+            self.config['InlineBackend']['rc'].update({_rc})
+        elif 'InlineBackend' in self.config:
+            self.config['InlineBackend'].update({'rc': _rc})
+        else:
+            self.config.update({'InlineBackend': Config({'rc': _rc})})
+        rc = self.config['InlineBackend']['rc']
+
+        # This seems to be necessary for newer versions of Traitlets because
+        # print_figure_kwargs doesn't return a dict.
+        if isinstance(rc, LazyConfigValue):
+            rc = rc.to_dict().get('update') or rc
+
+        self._set_config_option('InlineBackend.rc', rc)
 
     def set_sympy_forecolor(self, background_color='dark'):
         """Set SymPy forecolor depending on console background."""
--- a/spyder_kernels/console/kernel.py
+++ b/spyder_kernels/console/kernel.py
@@ -950,10 +950,13 @@ class SpyderKernel(IPythonKernel):
         """
         Update InlineBackend given an option and value.
 
-        As parameters:
-            option: config option; one of 'close_figures', 'figure_formats',
-                'print_figure_kwargs', or 'rc'.
-            value: value of the option
+        Parameters
+        ----------
+        option: str
+            Configuration option. One of 'close_figures', 'figure_formats',
+            'print_figure_kwargs', or 'rc'.
+        value: str | dict
+            Value of the option.
         """
         if (
             'InlineBackend' in self.config
@@ -976,10 +979,13 @@ class SpyderKernel(IPythonKernel):
         if option == 'rc' and self.get_matplotlib_backend() == 'inline':
             # Explicitly update rcParams if already in inline mode so that
             # new settings are effective immediately.
-            import matplotlib
-            matplotlib.rcParams.update(value)
+            try:
+                import matplotlib
+                matplotlib.rcParams.update(value)
+            except Exception:
+                pass
 
-    def _restore_rc_file_defaults(self):
+    def restore_rc_file_defaults(self):
         """Restore inline rcParams to file defaults"""
         try:
             import matplotlib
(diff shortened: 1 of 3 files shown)


spyder_kernels/console/kernel.py
```diff
@@ -540,67 +540,52 @@ def set_matplotlib_conf(self, conf):
...
-            self._set_mpl_inline_rc_config('font.size', conf[fontsize_n])
-
+            rc.update({'font.size': conf[fontsize_n]})
         if bottom_n in conf:
-            self._set_mpl_inline_rc_config(
-                'figure.subplot.bottom',
-                conf[bottom_n]
-            )
+            rc.update({'figure.subplot.bottom': conf[bottom_n]})
+        if rc:
```
ccordoba12, line 558:
```suggestion

        # Update Inline backend parameters, if available.
        if rc:
```

spyder_kernels/console/kernel.py
```diff
@@ -540,67 +540,52 @@ def set_matplotlib_conf(self, conf):
...
+        pylab_backend_o = conf.get(pylab_backend_n, current_backend)
+        backend_changed = current_backend != pylab_backend_o
+        if (
+            current_backend == 'inline' and rc
+            and not backend_changed
+            and interactive_backend not in ('inline', -1)
+        ):
+            self._set_mpl_backend(interactive_backend)
+        if (
+            (current_backend == 'inline' and rc)  # toggle back to inline
```
ccordoba12, line 584:
```suggestion

        # Toggle back to inline
        if (
            (current_backend == 'inline' and rc)
```

spyder_kernels/console/kernel.py
```diff
@@ -540,67 +540,52 @@ def set_matplotlib_conf(self, conf):
...
-            if figure_kwargs_dict:
-                print_figure_kwargs = figure_kwargs_dict
-
-        self._set_config_option(
-            'InlineBackend.print_figure_kwargs', print_figure_kwargs)
+        # To update rcParams in inline mode we can either do so directly or
+        # re-assert inline mode. However, either will prevent restoring
+        # rcParams when toggling to interactive mode. The workaround is to
+        # first toggle to interactive mode, then back to inline. This updates
+        # the rcParams and restores rcParams when switching to interactive.
```
ccordoba12, line 571: Unfortunately, this is a hack that wouldn't allow users to set their preferred interactive backend using the `%matplotlib` magic. And that would break the workflow of several power users (including myself).

The problem is that with your approach we'd be auto-setting the interactive backend for users with the options saved in Spyder. But we shouldn't be doing that because they can do it with `%matplotlib` too (and you know that once an interactive backend is set in a session, it can't be changed).

Is it not possible to change Matplotlib rc parameters after switching to an interactive backend? Or is the issue that Matplotlib only allows to set rc parameters once per session? (A quick googling [seems to suggest](https://matplotlib.org/stable/api/matplotlib_configuration_api.html#matplotlib.rcdefaults) that rc parameters can be changed on the fly).

spyder_kernels/console/kernel.py
```diff
@@ -958,16 +943,31 @@ def _set_config_option(self, option, value):
...
+            'InlineBackend' in self.config
+            and option in self.config['InlineBackend']
+            and isinstance(value, dict)
+        ):
+            self.config['InlineBackend'][option].update(value)
+        elif 'InlineBackend' in self.config:
+            self.config['InlineBackend'].update({option: value})
+        else:
+            self.config.update({'InlineBackend': {option: value}})
+        value = self.config['InlineBackend'][option]
```
ccordoba12, line 965:
```suggestion

        value = self.config['InlineBackend'][option]
```

ccordoba12 (commented):
@mrclary, sorry to say it but I think your current approach is not feasible (see below).

spyder_kernels/console/kernel.py
mrclary: Toggling back to inline is only one of the reasons to run the code in this `if` block, not the only one. The other reasons are if autoloading pylab or the backend has changed.

spyder_kernels/console/kernel.py
mrclary: > Unfortunately, this is a hack that wouldn't allow users to set their preferred interactive backend using the `%matplotlib` magic. And that would break the workflow of several power users (including myself).
> 
> The problem is that with your approach we'd be auto-setting the interactive backend for users with the options saved in Spyder. But we shouldn't be doing that because they can do it with `%matplotlib` too (and you know that once an interactive backend is set in a session, it can't be changed).

Aahh, yes, that is a problem...

> Is it not possible to change Matplotlib rc parameters after switching to an interactive backend? Or is the issue that Matplotlib only allows to set rc parameters once per session? (A quick googling [seems to suggest](https://matplotlib.org/stable/api/matplotlib_configuration_api.html#matplotlib.rcdefaults) that rc parameters can be changed on the fly).

It is possible to set rc parameters after switching to an interactive backend and they can be changed on the fly.

The principle issue is two-fold:
* Manually setting `rcParams` while in inline mode forces these parameters to persist after changing to an interactive backend, resulting unexpected values when switching between interactive and inline modes.
* `rcParams` inherit IPython `InlineBackend` settings only upon activating the inline backend, but re-activating the inline backend while in inline mode also forces these parameters to persist, resulting in unexpected values when switching between interactive and inline modes.

This hack only applies to the circumstance where the user changes Spyder's inline preferences while in inline mode. If I remove the hack, changes to Spyder's inline preferences would not be applied to existing kernels until after changing from, and back to, inline mode. This would not affect other expected functionality and would not preset the interactive backend.

The only other alternative is to explicitly set `rcParams` every time the backend changes: `matplotlib.rc_file_defaults()` on changing to interactive backends; apply Spyder's inline preferences on activating inline mode (or when preferences change while in inline mode). This would require storing the preferences in the `kernel` object and obviates the use of `InlineBackend` settings. I thought it was too heavy-handed and inelegant, but it can nevertheless be done.

spyder_kernels/console/kernel.py
mrclary: Nevertheless, this code block will certainly change given the comments above regarding the toggling "hack".

spyder_kernels/console/kernel.py
ccordoba12: > The only other alternative is to explicitly set rcParams every time the backend changes

Yep, and I think we should do precisely that. Perhaps it's not as elegant as your current approach but it's more explicit and less hacky.

mrclary: Actually, I don't think this will work either. I can only explicitly set rcParams in spyder-kernels, and toggling the backend via the status bar just sends the magic command directly to the console, which would bypass any mechanism in the kernel to set the rcParam. Furthermore, even if I find a way around this, a user can still explicitly use a magic in the console and bypass the mechanism. Therefore, there will always be an inconsistent update to rcParams.

I think the only way to have consistent, repeatable application of the rcParams is for Spyder to not explicitly set them and rely only on the InlineBackend settings mechanism and accept that changing the inline preferences will not have immediate effect if already in inline mode.

ccordoba12: > Actually, I don't think this will work either. I can only explicitly set rcParams in spyder-kernels, and toggling the backend via the status bar just sends the magic command directly to the console, which would bypass any mechanism in the kernel to set the rcParam.

Yeah, but the nice thing is that we're already overriding the command that's run when calling `%matplotlib`:

https://github.com/spyder-ide/spyder-kernels/blob/94d107e4fecf4a1b910feb41b96e6f228ab3f9c1/spyder_kernels/console/shell.py#L87-L92

So, you should be able to switch rcParams between inline and interactive before actually setting the backend. And since the shell has access to the kernel (through `self.kernel`) you can leave the methods to do that in the kernel so that they work both when receiving new options from Spyder and using the magic.

mrclary: I didn't realize we were overriding `%matplotlib`; I think that will work perfectly for this purpose. I pushed up new changes.

spyder_kernels/console/kernel.py
```diff
@@ -972,16 +946,54 @@ def _set_config_option(self, option, value):
...
-    def _set_mpl_inline_rc_config(self, option, value):
+    def _set_inline_config_option(self, option, value):
         """
-        Update any of the Matplolib rcParams given an option and value.
+        Update InlineBackend given an option and value.
+
+        As parameters:
+            option: config option; one of 'close_figures', 'figure_formats',
+                'print_figure_kwargs', or 'rc'.
+            value: value of the option
```
ccordoba12, line 956: Please rewrite this part in the Numpy docstring format, i.e.

```python
"""
Parameters
----------
option: str
    ...
"""
```

spyder_kernels/console/kernel.py
```diff
@@ -972,16 +946,54 @@ def _set_config_option(self, option, value):
...
+        if isinstance(value, LazyConfigValue):
+            value = value.to_dict().get('update') or value
+
+        self._set_config_option(f'InlineBackend.{option}', value)
+
+        if option == 'rc' and self.get_matplotlib_backend() == 'inline':
+            # Explicitly update rcParams if already in inline mode so that
+            # new settings are effective immediately.
+            import matplotlib
+            matplotlib.rcParams.update(value)
```
ccordoba12, line 980:
```suggestion
            try:
                import matplotlib
                matplotlib.rcParams.update(value)
            except Exception:
                pass
```

Just in case Matplotlib fails to import for whatever reason.

spyder_kernels/console/kernel.py
```diff
@@ -972,16 +946,54 @@ def _set_config_option(self, option, value):
...
+
+    def _restore_rc_file_defaults(self):
```
ccordoba12, line 982:
```suggestion
    def restore_rc_file_defaults(self):
```

This needs to be a public method because it's used outside this class.

spyder_kernels/console/shell.py
```diff
@@ -89,6 +89,10 @@ def enable_matplotlib(self, gui=None):
         if gui is None or gui.lower() == "auto":
             gui = automatic_backend()
 
+        # Before activating the backend, restore to file default those
+        # InlineBackend settings that may have been set explicitly.
+        self.kernel._restore_rc_file_defaults()
```
ccordoba12, line 94:
```suggestion
        self.kernel.restore_rc_file_defaults()
```

To go with my suggestion above.

ccordoba12 (commented):
This looks much better, thanks @mrclary!

spyder_kernels/console/kernel.py
mrclary:
```suggestion
        Parameters
        ----------
        option: str
            Configuration option. One of 'close_figures', 'figure_formats',
            'print_figure_kwargs', or 'rc'.
        value: str | dict
            Value of the option. 
```

ccordoba12 (approved):
Looks good to me now, thanks @mrclary!

PR #488, opened 2024-05-20
Accommodate Matplotlib 3.9.0 in spyder-kernels 3.x

mrclary:
Matplotlib `get_backend` returns lowercase in version >=3.9.0.
Update `MPL_BACKENDS_TO_SPYDER` to accommodate.


ccordoba12:
This is not really necessary because the fix will be picked up when merging the `2.x` branch to `master` after merging PR #489.

But I guess it was necessary to test the fix worked in Spyder's `master` branch.

PR #490, opened 2024-05-23
PR: use inodes for single canonic file path

impact27:
Fixes: TODO


impact27:
@ccordoba12 the tests fail for 3.8 but I don't see any log? I just get a cross with "This job failed"

ccordoba12:
> the tests fail for 3.8 but I don't see any log? I just get a cross with "This job failed"

Probably Github actions was failing at the time. But I restarted them and they're working fine now.

ccordoba12:
About this PR: how hard do you think it could be to add a test based on the example you posted on issue #491?

impact27:
I'll add that

ccordoba12 (approved):

PR #499, opened 2024-08-26
PR: Resolve possible macOS application bundle symlink

mrclary:
If pyexec is not a symlink, nothing will change. If pyexec is a symlink and in a conda environment, then the symlink will resolve to pythonX.XX in the same directory, so the result of get_conda_env_path will be the same. If pyexec is a symlink that is neither in a conda environment nor in a macOS application bundle, then the result get_conda_env_path may be different, but it wasn't a conda environment anyway.


mrclary:
Closing as unnecessary.

PR #503, opened 2024-09-27
PR: Update `load_dicom` to accommodate Pydicom 3.0

mrclary:
`pydicom` released version 3.0 three weeks ago, resulting in recent `spyder-kernels` test failures for `spyder_kernels/utils/tests/test_iofuncs.py::test_load_dicom_files`.

Unfortunately, this includes breaking changes. In particular, `pydicom.dicomio.read_file` is renamed `pydicom.dicomio.dcmread`. Also unfortunate, `pydicom` does not support Python <3.10. Also unfortunate, this requirement is neither enforced by `pip` nor `conda` for some reason, i.e. `pydicom` 3.0.1 is installed into Python <3.10 environments.

* Update `load_dicom` to accommodate only two configurations
  - Pydicom >=3.0 and Python >=3.10
  - Pydicom <3.0 and Python <3.10
* Skip `test_load_dicom_files` if Python <3.10, since CI will install Pydicom >=3

### DIFF
--- a/spyder_kernels/utils/tests/test_iofuncs.py
+++ b/spyder_kernels/utils/tests/test_iofuncs.py
@@ -14,7 +14,6 @@ Tests for iofuncs.py.
 import copy
 import io
 import os
-import sys
 
 # Third party imports
 from PIL import ImageFile
@@ -352,8 +351,8 @@ def test_save_load_hdf5_files(tmp_path):
 
 
 @pytest.mark.skipif(
-    sys.version_info < (3, 10, 0),
-    reason="pydicom 3.0 is incompatible with Python < 3.10"
+    os.environ.get("USE_CONDA") == "true",
+    reason="Pydicom is not installed correctly in Conda envs"
 )
 def test_load_dicom_files():
     """Check that we can load DICOM files."""
--- a/spyder_kernels/utils/tests/test_iofuncs.py
+++ b/spyder_kernels/utils/tests/test_iofuncs.py
@@ -11,9 +11,10 @@ Tests for iofuncs.py.
 """
 
 # Standard library imports
+import copy
 import io
 import os
-import copy
+import sys
 
 # Third party imports
 from PIL import ImageFile
@@ -350,6 +351,10 @@ def test_save_load_hdf5_files(tmp_path):
     assert repr(iofuncs.load_hdf5(h5_file)) == repr(expected)
 
 
+@pytest.mark.skipif(
+    sys.version_info < (3, 10, 0),
+    reason="pydicom 3.0 is incompatible with Python < 3.10"
+)
 def test_load_dicom_files():
     """Check that we can load DICOM files."""
     # This test pass locally but we need to set the variable below for it to
--- a/requirements/tests.txt
+++ b/requirements/tests.txt
@@ -12,4 +12,4 @@ xarray
 pillow
 django
 h5py
-pydicom
+pydicom<3.0
--- a/requirements/tests.txt
+++ b/requirements/tests.txt
@@ -12,4 +12,4 @@ xarray
 pillow
 django
 h5py
-pydicom<3.0
+pydicom
--- a/requirements/tests.txt
+++ b/requirements/tests.txt
@@ -12,4 +12,4 @@ xarray
 pillow
 django
 h5py
-pydicom
+pydicom>=3.0
--- a/requirements/tests.txt
+++ b/requirements/tests.txt
@@ -12,4 +12,4 @@ xarray
 pillow
 django
 h5py
-pydicom>=3.0
+pydicom
--- a/setup.py
+++ b/setup.py
@@ -61,7 +61,7 @@ TEST_REQUIREMENTS = [
     'pillow',
     'django',
     'h5py',
-    'pydicom'
+    'pydicom>=3.0'
 ]
 
 setup(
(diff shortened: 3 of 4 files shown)


mrclary:
@ccordoba12, I think that the environment cache will need to be cleared so that the Linux pip test for Python 3.10 will pass. `pydicom` 3.0.1 is still being installed instead of 2.4.4.

ccordoba12:
> Unfortunately, this includes breaking changes. In particular, pydicom.dicomio.read_file is renamed pydicom.dicomio.dcmread.

Ok, then, please change this

```python
try:
    data = dicomio.read_file(filename, force=True)
except TypeError:
    data = dicomio.read_file(filename)
```

to be

```python
try:
   # For Pydicom 3/Python 3.10+
    data = dicomio.dcmread(filename, force=True)
except TypeError:
    data = dicomio.dcmread(filename)
except AttributeError:
    # For Pydicom 2/Python 3.9-
    try:
        data = dicomio.read_file(filename, force=True)
    except TypeError:
        data = dicomio.read_file(filename)
```

That will make that code run for Pydicom greater than and less than 3.0.

> Also unfortunate, pydicom does not support Python <3.10. Also unfortunate, this requirement is neither enforced by pip nor conda for some reason, i.e. pydicom 3.0.1 is installed into Python <3.10 environments.

Ok, then please skip the test that checks our Pydicom integration for Python versions less than 3.10. And then you can remove the constraint you added to Pydicom.

It'd be nice too if you could inform Pydicom and Conda-forge maintainers that it's 3.0 version only support 3.10+. But that's up to you.

ccordoba12:
> Also unfortunate, this requirement is neither enforced by pip nor conda for some reason, i.e. pydicom 3.0.1 is installed into Python <3.10 environments.

The requirement is correctly enforced by pip. Otherwise our tests with pip packages would fail with the last change you did in this PR. So, the problem is only present in Conda-forge.

And in that case, I think the best we can do is to submit a PR to the Pydicom feedstock to fix the situation.

mrclary:
> ```python
> try:
>    # For Pydicom 3/Python 3.10+
>     data = dicomio.dcmread(filename, force=True)
> except TypeError:
>     data = dicomio.dcmread(filename)
> except AttributeError:
>     # For Pydicom 2/Python 3.9-
>     try:
>         data = dicomio.read_file(filename, force=True)
>     except TypeError:
>         data = dicomio.read_file(filename)
> ```
> 
> That will make that code run for Pydicom greater than and less than 3.0.
> 
> > Also unfortunate, pydicom does not support Python <3.10. Also unfortunate, this requirement is neither enforced by pip nor conda for some reason, i.e. pydicom 3.0.1 is installed into Python <3.10 environments.
> 
> Ok, then please skip the test that checks our Pydicom integration for Python versions less than 3.10. And then you can remove the constraint you added to Pydicom.

If we skip Pydicom integration test for Python <3.10, then there is no need catch the `AttributeError`.

ccordoba12:
> If we skip Pydicom integration test for Python <3.10, then there is no need catch the AttributeError

This is not only a matter of fixing our tests. It also has to do with maintaining Pydicom support for Python 3.8/3.9 and 3.10+ at the same time. We need to do that because 3.9 will reach end-of-life in a year or so (3.8 will become unsupported at the end of October though).

mrclary:
> > If we skip Pydicom integration test for Python <3.10, then there is no need catch the AttributeError
> 
> This is not only a matter of fixing our tests. It also has to do with maintaining Pydicom support for Python 3.8/3.9 and 3.10+ at the same time. We need to do that because 3.9 will reach end-of-life in a year or so (3.8 will become unsupported at the end of October though).

Yep, you are correct. I misspoke.

mrclary:
@ccordoba12, this is ready for review

mrclary:
Actually, `pydicom = 3.0.1_1` was just built per conda-forge/pydicom-feedstock#40. This should correctly enforce <3.0 for Python <3.10, so we may not have to skip the test.

spyder_kernels/utils/tests/test_iofuncs.py
```diff
@@ -11,9 +11,10 @@
 """
 
 # Standard library imports
+import copy
 import io
 import os
-import copy
+import sys
```
ccordoba12, line 17:
```suggestion
```

This is not necessary (see below).

spyder_kernels/utils/tests/test_iofuncs.py
```diff
@@ -350,6 +351,10 @@ def test_save_load_hdf5_files(tmp_path):
     assert repr(iofuncs.load_hdf5(h5_file)) == repr(expected)
 
 
+@pytest.mark.skipif(
+    sys.version_info < (3, 10, 0),
+    reason="pydicom 3.0 is incompatible with Python < 3.10"
+)
```
ccordoba12, line 357:
```suggestion
@pytest.mark.skipif(
    os.environ.get("USE_CONDA") == "true",
    reason="Pydicom is not installed correctly in Conda envs"
)
```

The test should run as expected with pip packages for Python 3.10+ and 3.9-, which is much better.

ccordoba12 (commented):
One additional suggestion so `test_load_dicom_files` works for all our supported Python versions, but only when using pip packages.

ccordoba12 (approved):
Looks good to me now, thanks @mrclary!

ccordoba12:

PR #513, opened 2024-10-04
PR: Handle internal  '_pdbcmd'

impact27:
Pdb added internal commands that starts with "_pdbcmd"

fixes https://github.com/spyder-ide/spyder/issues/22500


ccordoba12:
@impact27, I think the problem requires a more generic solution because more Pdb commands (besides `_pdbcmd_*` ones) could be added to Pdb's `cmdqueue` in the future (see https://github.com/python/cpython/pull/120928 for how that command was introduced). So, we need to process them as such instead of handling this particular case differently.

That's why I opened PR #530, which I verified fixes the problem too.

PR #517, opened 2024-10-11
PR: Manually register the Matplotlib inline backend in case it wasn't

ccordoba12:
- This can be necessary when there's a mismatch between the installed versions of IPython, matplotlib and matplotlib-inline.
- Addresses spyder-ide/spyder#22420


ccordoba12:

PR #520, opened 2024-11-13
Handle async do_complete in the debugger

Carreau:
A number of methods of ipykernel can optinally return `awaitable[T]` instead of just `T`, this is the case for `do_complete`.

I think it's a mistake ; see ipython/ipykernel#1295 ; in particular because it's easy to forget / hard to properly type-check, and I'd like to make it mandatory in the long term to have await.

Spyder seem to not handle the case where do_completer return an awaitable (or more partiularly is `do_complete` is a coroutine function.

This tries to handle it – and as of course `do_completer` _can_ be async, all caller _must_ be async. So I try to do all the required updates.

Note: I also add explict imports in some test, to get better error message in case those deps are not installed.

### DIFF
--- a/spyder_kernels/console/tests/test_console_kernel.py
+++ b/spyder_kernels/console/tests/test_console_kernel.py
@@ -901,13 +901,15 @@ def test_matplotlib_inline(kernel):
         assert 'inline' in value
 
 
-def test_do_complete(kernel):
+async def test_do_complete(kernel):
     """
     Check do complete works in normal and debugging mode.
     """
     asyncio.run(kernel.do_execute('abba = 1', True))
     assert kernel.get_value('abba') == 1
     match = kernel.do_complete('ab', 2)
+    if isawaitable(match):
+        match = await match
     assert 'abba' in match['matches']
 
     # test pdb
@@ -916,6 +918,8 @@ def test_do_complete(kernel):
     pdb_obj.completenames = lambda *ignore: ['baba']
     kernel.shell._namespace_stack = [pdb_obj]
     match = kernel.do_complete('ba', 2)
+    if isawaitable(match):
+        match = await match
     assert 'baba' in match['matches']
     pdb_obj.curframe = None
 
@@ -1435,6 +1439,7 @@ def test_django_settings(kernel):
 
     This is a regression test for issue spyder-ide/spyder#19516
     """
+    import django
     asyncio.run(kernel.do_execute('from django.conf import settings', True))
     nsview = repr(kernel.get_namespace_view())
     assert "'settings':" in nsview
--- a/spyder_kernels/utils/tests/test_iofuncs.py
+++ b/spyder_kernels/utils/tests/test_iofuncs.py
@@ -342,6 +342,7 @@ def test_spydata_export(input_namespace, expected_namespace,
 
 def test_save_load_hdf5_files(tmp_path):
     """Simple test to check that we can save and load HDF5 files."""
+    import h5py
     h5_file = tmp_path / "test.h5"
     data = {'a' : [1, 2, 3, 4], 'b' : 4.5}
     iofuncs.save_hdf5(data, h5_file)
--- a/spyder_kernels/customize/spyderpdb.py
+++ b/spyder_kernels/customize/spyderpdb.py
@@ -351,16 +351,16 @@ class SpyderPdb(ipyPdb):
     do_bt = do_where
 
     # --- Method defined by us to respond to ipython complete protocol
-    def do_complete(self, code, cursor_pos):
+    async def do_complete(self, code, cursor_pos):
         """
         Respond to a complete request.
         """
         if self.pdb_use_exclamation_mark:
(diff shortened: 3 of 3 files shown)


Carreau:
This is more to start a discussion than a "this must be merged". It's mostly to start some discussion also with upstream as we sort of disagree that having `def f(...): -> T | Awaitable[T]` is a good thing.

Carreau:
Rebased. I think we agree upstream that things should mostly be async when possible. so +1 to merge this once tests are passing, and you agree.,

spyder_kernels/console/tests/test_console_kernel.py
```diff
@@ -1435,6 +1439,7 @@ def test_django_settings(kernel):
 
     This is a regression test for issue spyder-ide/spyder#19516
     """
+    import django
```
ccordoba12, line 1442: Why is this necessary now? It wasn't before.

spyder_kernels/utils/tests/test_iofuncs.py
```diff
@@ -342,6 +342,7 @@ def test_spydata_export(input_namespace, expected_namespace,
 
 def test_save_load_hdf5_files(tmp_path):
     """Simple test to check that we can save and load HDF5 files."""
+    import h5py
```
ccordoba12, line 345: And this?

ccordoba12 (commented):
Thanks @Carreau! A couple of small questions, the rest looks good to me.

spyder_kernels/console/tests/test_console_kernel.py
ccordoba12: Ok, thanks for the clarification. Merging then.

ccordoba12 (approved):
Thanks @Carreau!

Carreau:
> But that's not actually the case in IPykernel's main nor in the `6.x` branch:
> 
> [ipython/ipykernel@`cac11a0`/ipykernel/ipkernel.py#L512-L532](https://github.com/ipython/ipykernel/blob/cac11a08449d41ceb8a56e3095e78e00cb1626eb/ipykernel/ipkernel.py#L512-L532)
> 
> So, I guess these changes are unnecessary and need to be reverted, right?

The code that _call_ do_complete (and similar) accept both T|awaitable[T] and emit a warning that returning awaitable[T] might become mandatory in the future.

https://github.com/ipython/ipykernel/blob/cac11a08449d41ceb8a56e3095e78e00cb1626eb/ipykernel/kernelbase.py#L830-L838

Some subclass of ipykernel (I don't remember which) use `async do_complete`, maybe xeus maybe another...

I haven't changed the base methods because it would break downstream (like Spyder kernel): https://github.com/ipython/ipykernel/pull/1339. This PR ensure that Spyder kernel is not broken _when_ we change the base methods, but we need a release cycle for that.

Is there any chances you run with warnings turned into errors and this breaks the completions ?

But if you want to revert, feel free to do so.

ccordoba12:
> The code that call do_complete (and similar) accept both T|awaitable[T] and emit a warning that returning awaitable[T] might become mandatory in the future.

Ok, but the thing is through this method:

https://github.com/spyder-ide/spyder-kernels/blob/2e19c2ca469d4813c7f04500600e8dd005d42ffc/spyder_kernels/customize/spyderpdb.py#L425

we're calling IPylkernel's `do_complete` directly:

https://github.com/spyder-ide/spyder-kernels/blob/2e19c2ca469d4813c7f04500600e8dd005d42ffc/spyder_kernels/console/kernel.py#L451-L453

And since it's not async today, we're getting this error:

```python-traceback
[SpyderKernelApp] ERROR | Exception in control handler:
  File "/home/carlos/miniconda3/envs/py310-pip/lib/python3.10/site-packages/ipykernel/kernelbase.py", line 362, in process_control
    await result
  File "/home/carlos/miniconda3/envs/py310-pip/lib/python3.10/site-packages/ipykernel/kernelbase.py", line 831, in complete_request
    matches = await matches
  File "/home/carlos/Projects/spyder/spyder/external-deps/spyder-kernels/spyder_kernels/customize/spyderpdb.py", line 359, in do_complete
    return await self._complete_exclamation(code, cursor_pos)
  File "/home/carlos/Projects/spyder/spyder/external-deps/spyder-kernels/spyder_kernels/customize/spyderpdb.py", line 532, in _complete_exclamation
    result = await self.shell.kernel._do_complete(code, cursor_pos)
TypeError: object dict can't be used in 'await' expression
```

When googling for that error, I found that `do_complete` in IPykernel should be async to fix it.

Or do you know another way in which we could solve this?

Carreau:
Ah, 

then you need to do like in this PR:

result = call()
if isawaitable(result):
    result = await result

this will handle both awaitable and non awaitable. This will work for both an async def function, or a sync function returning a future.

Carreau:
#534 should help, I haven't tried.

ccordoba12:
Thanks @Carreau! That fixes completions in our debugger with some minor tweaks.

PR #521, opened 2024-11-16
PR: Add logic to handle traceback color configuration

dalthviz:
Part of spyder-ide/spyder#22412

### DIFF
--- a/spyder_kernels/console/kernel.py
+++ b/spyder_kernels/console/kernel.py
@@ -639,6 +639,8 @@ class SpyderKernel(IPythonKernel):
                     ret["special_kernel_error"] = value
             elif key == "color scheme":
                 self.set_color_scheme(value)
+            elif key == "traceback_highlight_style":
+                self.set_traceback_syntax_highlighting(value)
             elif key == "jedi_completer":
                 self.set_jedi_completer(value)
             elif key == "greedy_completer":
@@ -664,6 +666,25 @@ class SpyderKernel(IPythonKernel):
         elif color_scheme == "light":
             self.shell.run_line_magic("colors", "lightbg")
             self.set_sympy_forecolor(background_color='light')
+
+    def set_traceback_syntax_highlighting(self, syntax_style):
+        """Set the traceback syntax highlighting."""
+        try:
+            import IPython.core.ultratb
+            from IPython.core.ultratb import VerboseTB
+
+            from spyder.plugins.ipythonconsole.utils.style import create_style_class
+
+            IPython.core.ultratb.get_style_by_name = create_style_class
+
+            if getattr(VerboseTB, 'tb_highlight_style', None) is not None:
+                VerboseTB.tb_highlight_style = syntax_style
+            elif getattr(VerboseTB, '_tb_highlight_style', None) is not None:
+                VerboseTB._tb_highlight_style = syntax_style
+        except Exception:
+            # Only usable if the kernel has access to Spyder syntax style function logic
+            # i.e spyder-kernels and Spyder are installed in the same environment
+            pass
 
     def get_cwd(self):
         """Get current working directory."""
--- a/spyder_kernels/console/kernel.py
+++ b/spyder_kernels/console/kernel.py
@@ -47,6 +47,7 @@ from spyder_kernels.utils.iofuncs import iofunctions
 from spyder_kernels.utils.mpl import automatic_backend, MPL_BACKENDS_TO_SPYDER
 from spyder_kernels.utils.nsview import (
     get_remote_data, make_remote_view, get_size)
+from spyder_kernels.utils.style import create_style_class
 from spyder_kernels.console.shell import SpyderShell
 from spyder_kernels.comms.utils import WriteContext
 
@@ -659,17 +660,13 @@ class SpyderKernel(IPythonKernel):
         return ret
 
     def set_color_scheme(self, color_scheme):
-        if color_scheme == "dark":
-            # Needed to change the colors of tracebacks
-            self.shell.run_line_magic("colors", "linux")
-        elif color_scheme == "light":
-            self.shell.run_line_magic("colors", "lightbg")
+        self.shell.set_spyder_theme(color_scheme)
         self.set_sympy_forecolor(background_color=color_scheme)
         self.set_traceback_highlighting(color_scheme)
 
     def set_traceback_highlighting(self, color_scheme):
         """Set the traceback highlighting color."""
-        color = 'bg:ansigreen' if color_scheme == 'dark' else 'bg:ansiyellow'
+        color = 'bg:ansired' if color_scheme == 'dark' else 'bg:ansiyellow'
         from IPython.core.ultratb import VerboseTB
 
         if getattr(VerboseTB, 'tb_highlight', None) is not None:
@@ -681,8 +678,6 @@ class SpyderKernel(IPythonKernel):
         """Set the traceback syntax highlighting style."""
         import IPython.core.ultratb
         from IPython.core.ultratb import VerboseTB
-
-        from spyder_kernels.utils.style import create_style_class
 
         IPython.core.ultratb.get_style_by_name = create_style_class
 
--- a/spyder_kernels/console/kernel.py
+++ b/spyder_kernels/console/kernel.py
@@ -669,7 +669,7 @@ class SpyderKernel(IPythonKernel):
 
     def set_traceback_highlighting(self, color_scheme):
         """Set the traceback highlighting color."""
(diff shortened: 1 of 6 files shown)


spyder_kernels/console/kernel.py
```diff
@@ -660,10 +662,39 @@ def set_color_scheme(self, color_scheme):
...
+        elif getattr(VerboseTB, '_tb_highlight', None) is not None:
+            VerboseTB._tb_highlight = color
+
+    def set_traceback_syntax_highlighting(self, syntax_style):
+        """Set the traceback syntax highlighting style."""
+        try:
+            import IPython.core.ultratb
+            from IPython.core.ultratb import VerboseTB
+
+            from spyder.plugins.ipythonconsole.utils.style import create_style_class
```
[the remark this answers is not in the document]
ccordoba12: > Should the logic related with the creation of a Pygments style class while Spyder is running be moved or copied to spyder-kernels @ccordoba12 ?

To be able to support this in a Spyder bugfix release (i.e. 6.0.3), that code would need to be copied to Spyder-kernels (otherwise we'd need to require people to update spyder-kernels to 3.0.2 in all their envs). Then, in 6.1 we could remove that code from Spyder and import it directly from spyder-kernels.

> Maybe a package with those utility functions should be created?

I don't think that's necessary, i.e. having that code in spyder-kernels should be enough because Spyder depends directly on spyder-kernels.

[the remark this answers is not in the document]
ccordoba12: But I think you'd need to send to the kernel the result of the `get_color_scheme` function (in `spyder.config.gui`). In other words, that function shouldn't be moved and it shouldn't be hard to send its result to the kernel because it's probably a dict.

spyder_kernels/customize/umr.py
```diff
@@ -92,7 +92,7 @@ def run(self):
         # Report reloaded modules
         if self.verbose and modnames_to_reload:
             modnames = modnames_to_reload
-            print("\x1b[4;33m%s\x1b[24m%s\x1b[0m"
+            print("\x1b[1;4;31m%s\x1b[24m%s\x1b[0m"
```
ccordoba12, line 95: In order to make this color configurable according to the light/dark theme, you need to make `UserModuleReloader` receive `shell` as a kwarg, then pass `self.shell` to it here:

https://github.com/spyder-ide/spyder-kernels/blob/4e554db2ccf2bafa30a6dd2c5c74213e013efa0c/spyder_kernels/customize/code_runner.py#L148-L150

and finally add two methods to `SpyderShell` called `set_spyder_theme`/`get_spyder_theme` to set and get the current theme in a private variable. You'll need to call `set_spyder_theme` from `SpyderKernel` (when you get the theme info from Spyder) and `get_spyder_theme` here to adjust the color.

spyder_kernels/utils/style.py
ccordoba12, line 1: Please format this file with Black to make it easier to read and after applying my suggestions below.

spyder_kernels/utils/style.py
```diff
@@ -0,0 +1,111 @@
...
+        else:
+            return ''
+
+    color_scheme = color_scheme_dict
+
+    fon_c, fon_fw, fon_fs = color_scheme['normal']
+    font_color =  fon_c
+    font_font_weight = give_font_weight(fon_fw)
+    font_font_style = give_font_style(fon_fs)
+    key_c, key_fw, key_fs = color_scheme['keyword']
```
ccordoba12, line 43:
```suggestion

    key_c, key_fw, key_fs = color_scheme['keyword']
```

For readability.

spyder_kernels/utils/style.py
```diff
@@ -0,0 +1,111 @@
...
+    keyword_color =  key_c
+    keyword_font_weight = give_font_weight(key_fw)
+    keyword_font_style = give_font_style(key_fs)
+    bui_c, bui_fw, bui_fs = color_scheme['builtin']
```
ccordoba12, line 47:
```suggestion

    bui_c, bui_fw, bui_fs = color_scheme['builtin']
```

spyder_kernels/utils/style.py
```diff
@@ -0,0 +1,111 @@
...
+    builtin_color =  bui_c
+    builtin_font_weight = give_font_weight(bui_fw)
+    builtin_font_style = give_font_style(bui_fs)
+    str_c, str_fw, str_fs = color_scheme['string']
```
ccordoba12, line 51:
```suggestion

    str_c, str_fw, str_fs = color_scheme['string']
```

spyder_kernels/utils/style.py
```diff
@@ -0,0 +1,111 @@
...
+    string_color =  str_c
+    string_font_weight = give_font_weight(str_fw)
+    string_font_style = give_font_style(str_fs)
+    num_c, num_fw, num_fs = color_scheme['number']
```
ccordoba12, line 55:
```suggestion

    num_c, num_fw, num_fs = color_scheme['number']
```

spyder_kernels/utils/style.py
```diff
@@ -0,0 +1,111 @@
...
+    number_color =  num_c
+    number_font_weight = give_font_weight(num_fw)
+    number_font_style = give_font_style(num_fs)
+    com_c, com_fw, com_fs = color_scheme['comment']
```
ccordoba12, line 59:
```suggestion

    com_c, com_fw, com_fs = color_scheme['comment']
```

spyder_kernels/utils/style.py
```diff
@@ -0,0 +1,111 @@
...
+    comment_color =  com_c
+    comment_font_weight = give_font_weight(com_fw)
+    comment_font_style = give_font_style(com_fs)
+    def_c, def_fw, def_fs = color_scheme['definition']
```
ccordoba12, line 63:
```suggestion

    def_c, def_fw, def_fs = color_scheme['definition']
```

spyder_kernels/utils/style.py
```diff
@@ -0,0 +1,111 @@
...
+    definition_color =  def_c
+    definition_font_weight = give_font_weight(def_fw)
+    definition_font_style = give_font_style(def_fs)
+    ins_c, ins_fw, ins_fs = color_scheme['instance']
```
ccordoba12, line 67:
```suggestion

    ins_c, ins_fw, ins_fs = color_scheme['instance']
```

ccordoba12 (commented):
Thanks @dalthviz for this great improvement!

spyder_kernels/utils/style.py
dalthviz: As a side note, should `pre-commit` with a black or ruff hook be added to this repo?

ccordoba12: Yep, I think that's a good idea.

spyder_kernels/customize/code_runner.py
```diff
@@ -145,10 +145,11 @@ class SpyderCodeRunner(Magics):
     def __init__(self, *args, **kwargs):
         self.show_global_msg = True
         self.show_invalid_syntax_msg = True
+        super().__init__(*args, **kwargs)
```
ccordoba12, line 148:
```suggestion
        super().__init__(*args, **kwargs)

        self.show_global_msg = True
        self.show_invalid_syntax_msg = True
```

It should be possible to call `super().__init__` at the beginning of this method, which is customary and makes the code easier to read.

spyder_kernels/customize/umr.py
```diff
@@ -92,7 +93,11 @@ def run(self):
...
         if self.verbose and modnames_to_reload:
             modnames = modnames_to_reload
-            print("\x1b[4;33m%s\x1b[24m%s\x1b[0m"
-                  % ("Reloaded modules", ": "+", ".join(modnames)))
+            colors = {"dark": "33", "light": "31"}
+            color = colors["dark"]
+            if self._shell:
+                color = colors[self._shell.get_spyder_theme()]
+            print("\x1b[4;%sm%s\x1b[24m%s\x1b[0m"
+                  % (color, "Reloaded modules", ": "+", ".join(modnames)))
```
ccordoba12, line 101: Please transform this to an f-string so it's easier to read.

ccordoba12 (commented):
Last couple of comments @dalthviz, then this should be ready.

spyder_kernels/utils/style.py
dalthviz: Created https://github.com/spyder-ide/spyder-kernels/issues/522 to track that task

ccordoba12 (approved):
Looks good to me now, thanks @dalthviz!

ccordoba12:

PR #526, opened 2025-01-16
PR: Filter frames that come from Spyder-kernels in tracebacks

ccordoba12:
- We don't need to show those frames to users because they contain no relevant info that they need to look at.
- This is more necessary now that we use  verbose tracebacks because those frames make a regular traceback much longer.


ccordoba12:

PR #527, opened 2025-01-16
PR: Use an older Ubuntu version on CIs

ccordoba12:
That version has the right system packages we need to run our tests.


ccordoba12:

PR #532, opened 2025-02-12
PR: Split `test_interrupt` into two tests

Carreau:
This is to know if only one or both are failing.
This also properly mark a test as skipped on Windows.

I guess we could make it a parametrize call, but
this would make it complicated to parametrize.


Carreau:
This is one of the ipykernel downstream failing tests.

PR #533, opened 2025-03-07
PR Add pixi related utility functions and env info

dalthviz:
Part of spyder-ide/spyder#23558

### DIFF
--- a/spyder_kernels/utils/pythonenv.py
+++ b/spyder_kernels/utils/pythonenv.py
@@ -19,6 +19,7 @@ class PythonEnvType:
     """Enum with the different types of Python environments we can detect."""
 
     Conda = "conda"
+    Pixi = "pixi"
     PyEnv = "pyenv"
     Custom = "custom"  # Nor Conda or Pyenv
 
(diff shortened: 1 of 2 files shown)


spyder_kernels/utils/pythonenv.py
```diff
@@ -61,6 +62,35 @@ def get_conda_env_path(pyexec, quote=False):
     return conda_env
 
 
+def get_pixi_manifest_path_and_env_name(pyexec, quote=False):
```
ccordoba12, line 65: Let's use `pathlib` instead of `os.path` for the path manipulations in this function because the logic is easier to follow.

ccordoba12 (commented):
Just a single suggestion for you @dalthviz.

spyder_kernels/utils/pythonenv.py
```diff
@@ -61,6 +62,33 @@ def get_conda_env_path(pyexec, quote=False):
...
+    pyexec_path = Path(pyexec.replace("\\", "/"))
+    pixi_env_path = pyexec_path.parent
+    pixi_env_name = pixi_env_path.name
+    pixi_dir_path = pixi_env_path.parent.parent
```
ccordoba12, line 69:
```suggestion
    pixi_dir_path = pixi_env_path.parents[1]
```

This is a simpler way to access the parent's parent.

ccordoba12:
@dalthviz, before merging: did you check that Pixi envs activation is working in Spyder after my suggestions?

ccordoba12 (approved):
Looks good to me now, thanks @dalthviz!

PR #534, opened 2025-03-13
PR: Fix some async do completes

Carreau:
Some versions of pytest do auto run async tests,
recents one do not, and marking with
`@pytest.mark.anyio` also test with multiple
runner if installed.

Handle case where do_complete is both async and
sync, and make _do_complete always async.

I only found mentions of the private _do_complete
in spyder kernel, so if present somewhere else you
might need to lift the if awaitable one level up.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,0 +1,16 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright © Spyder Kernels Contributors
+# Licensed under the terms of the MIT License
+#
+
+"""
+Main configuration file for Pytest
+"""
+
+import pytest
+
+
+@pytest.fixture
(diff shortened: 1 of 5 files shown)


ccordoba12:
@Carreau, I'll take care of this one to finish and merge it.

Carreau:
Thanks !

WHile at it, I don't know if you are registered for tidelift, but spyder-kernels seem to have some funds there:

https://tidelift.com/lifter/search/pypi/spyder-kernels

ccordoba12:
I didn't know that! Thanks for the heads up.

ccordoba12 (approved):
Thanks for your help to solve this @Carreau!

PR #535, opened 2025-03-31
PR: Display contents of string arrays

jitseniesen:
When displaying a NumPy array of strings, use `str` to convert the array to a string instead of using the default description "ndarray object of numpy module".

@ccordoba12 You turned off displaying the contents of string arrays in PR spyder-ide/spyder#5028 where you specified that only the contents of numeric arrays should be displayed. I presume that you did so to prevent problems with displaying arbitrary arrays, but I think string arrays are safe.

This fixes issue spyder-ide/spyder#22570.


ccordoba12:
> It looks like the CI script needs to be updated

Fixed in PR #538, please rebase.

jitseniesen:
Rebased and all tests pass :)

PR #536, opened 2025-04-01
PR: Preserve case sensitivity of working directory when running files

jsbautista:


spyder_kernels/customize/code_runner.py
```diff
@@ -286,6 +286,8 @@ def _exec_file(
             self.shell.showtraceback(exception_only=True)
             return
 
+        original_filename = filename
```
ccordoba12, line 289:
```suggestion
        # We need to keep the original file name to use it when setting the
        # working directory below.
        original_filename = filename
```

spyder_kernels/customize/code_runner.py
```diff
@@ -313,7 +315,7 @@ def _exec_file(
             if wdir is not None:
                 if wdir is True:
                     # True means use file dir
```
ccordoba12, line 317:
```suggestion
                # True means use file dir
                if wdir is True:
```

spyder_kernels/customize/code_runner.py
```diff
@@ -313,7 +315,7 @@ def _exec_file(
...
-                    wdir = os.path.dirname(filename)
+                    wdir = os.path.dirname(original_filename)
```
ccordoba12, line 318:
```suggestion
                    # We use the original file name to get the working
                    # directory in order to preserve its case sensitivity.
                    # Fixes spyder-ide/spyder#23835
                    wdir = os.path.dirname(original_filename)
```

ccordoba12 (commented):
Just some clarifying comments for you @jsbautista.

ccordoba12 (approved):
Thanks @jsbautista!

ccordoba12:

PR #541, opened 2025-05-01
PR: Support pathlib and None in namespace view

jitseniesen:
Support variables of type `pathlib.Path` and variables containing `None` in `value_to_display()`, so that they can be displayed in Spyder's Variable Explorer. Also add tests for this.

Variables of type `pathlib.Path` are converted using `str()`. The alternative would be to use `repr()`, which results in the variable `path` in the example below to be displayed as `PosixPath('/home/jitse/spyder-kernels')`. As explained in https://github.com/spyder-ide/spyder/issues/22351#issue-2467604479:

> The repr has the advantage as it gives a clear indicator in how to instantiate a new Path.
>
>  However displaying the printed version of the str form, where the escape characters are processed has the advantage in that it matches what is typically displayed in the Operating System and is consistent to the file paths seen elsewhere in Spyder

I went with `str()` because that is what we do in the Variable Explorer for strings and arrays.

Screenshot of Variable Explorer after the PR:

after-pathlib

Screenshot of current version, before the PR:

before-pathlib

### DIFF
--- a/spyder_kernels/utils/nsview.py
+++ b/spyder_kernels/utils/nsview.py
@@ -11,6 +11,7 @@ Utilities to build a namespace view.
 """
 from itertools import islice
 import inspect
+import pathlib
 import re
 
 from spyder_kernels.utils.lazymodules import (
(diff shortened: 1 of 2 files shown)


ccordoba12 (approved):
Looks good to me, thanks @jitseniesen!

PR #544, opened 2025-05-20
PR: Add minimal required version of `traitlets`

ccordoba12:
- Versions between 5.12 and 5.14.2 have a nasty bug that prevents setting the Matplotlib inline backend formats.
- This is necessary to address spyder-ide/spyder#24390


ccordoba12:

PR #549, opened 2025-06-08
PR: Support Polars dataframes

jitseniesen:
This PR adds support for displaying Polars data frames in Spyder's variable explorer to the Spyder kernels. The type is shown as "Polars DataFrame" and the value as "Column names: xxx" (same as for Pandas data frames). When the user opens an editor, the Polars data frame is converted to a Pandas data frame and sent to Spyder. 

This conversion fails if Pandas is not installed in the console environment. To handle this properly on the Spyder side, the last commit extends the error handling in remote calls to also send the `name` attribute (if it exists) of exceptions raised when handling remote calls. This attribute contains the name of the module whose import failed in `ModuleNotFoundError`.

This PR is used in Spyder PR spyder-ide/spyder#24558. See there for a video of how it all looks.

### DIFF
--- a/spyder_kernels/utils/nsview.py
+++ b/spyder_kernels/utils/nsview.py
@@ -564,7 +564,7 @@ def get_human_readable_type(item):
             if text == 'polars.dataframe.frame.DataFrame':
                 return 'Polars DataFrame'
             else:
-                return text[text.find('.')+1:]
+                return text[text.find('.') + 1:]
     except Exception:
         return 'Unknown'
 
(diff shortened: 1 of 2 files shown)


spyder_kernels/utils/nsview.py
```diff
@@ -558,7 +561,10 @@ def get_human_readable_type(item):
             return "Image"
         else:
             text = get_type_string(item)
-            return text[text.find('.')+1:]
+            if text == 'polars.dataframe.frame.DataFrame':
+                return 'Polars DataFrame'
+            else:
+                return text[text.find('.')+1:]
```
ccordoba12, line 567:
```suggestion
                return text[text.find('.') + 1:]
```

Minor code style improvement.

ccordoba12 (commented):
Just a few minor suggestions for you @jitseniesen.

jitseniesen:
> Just a few minor suggestions for you

I applied the suggestions and rebased, I think this is ready now.

PR #553, opened 2025-07-19
PR: Fix bug when converting JSON to StackSummary

jitseniesen:
This PR fixes what I believe to be a bug in `stacksummary_from_json()`. This function does indeed create a StackSummary, but it does not return it. As far as I can see, the return value is not really used anywhere, which is probably why nobody noticed. I want to use this information in the fix for issue spyder-ide/spyder#22411 and discovered the bug in the process.

This PR also adds a test and it fixes a typo in the function name.


spyder_kernels/comms/tests/test_cammbase.py
ccordoba12, line 1: I thinks this file should be named `test_commbase.py`.

ccordoba12 (commented):
Just a small comment for you @jitseniesen.

ccordoba12 (approved):
Thanks @jitseniesen!

PR #575, opened 2026-01-30
PR: Fix debugger stop in breakpoints set on modules if stop on first line option is disabled

rear1019:
This PR addresses spyder-ide/spyder#22025. This is not the fix we have been using internally for some time. Looking at it again I came up with a simpler and (hopefully) better fix.

The commit has a detailed explanation. It's … somewhat convoluted due to many layers being involved (Python's `pdb.Pdb`/`bdb.Bdb` and Spyder's customizations).

Please verify the fix, maybe @impact27 can give a hand. Unfortunately, I won't able to respond in the next two weeks to come.


ccordoba12 (approved):
Thanks for your help with this @rear1019!

Note: The failure in our tests is unrelated to this and I fixed it in PR #576.

ccordoba12:

PR #591, opened 2026-06-07
PR: Use generic args/kwargs in `SpyderPdb` init method

ccordoba12:
- That fixes an error when doing `breakpoint()` calls in Python 3.14.
- It also prevents issues in the future when args are changed in super classes.
- Also, start testing with Python 3.14.


ccordoba12:

PR #595, opened 2026-06-14
PR: Increase min Python version tested on Windows to 3.10 on CI

ccordoba12:
A single test started to fail on 3.9 but it couldn't be possible to fix it because that version reached end-of-life last year.


ccordoba12:

PR #600, opened 2026-07-15
PR: Add other Qt bindings to automatic backend detection

larsoner:
Updates logic to be more inclusive when deciding between `qt` and `tk`. Useful for https://github.com/spyder-ide/spyder/pull/25422

### DIFF
--- a/spyder_kernels/utils/mpl.py
+++ b/spyder_kernels/utils/mpl.py
@@ -31,7 +31,10 @@ MPL_BACKENDS_TO_SPYDER = {
 
 def automatic_backend():
     """Get Matplolib automatic backend option."""
-    if is_module_installed('PyQt5'):
+    if any(
+        is_module_installed(qt_binding)
+        for qt_binding in ('PyQt5', 'PySide2', 'PyQt6', 'PySide6')
+    ):
         auto_backend = 'qt'
     elif is_module_installed('_tkinter'):
         auto_backend = 'tk'


spyder_kernels/utils/mpl.py
```diff
@@ -31,7 +31,10 @@
 
 def automatic_backend():
     """Get Matplolib automatic backend option."""
-    if is_module_installed('PyQt5'):
+    if any(
+        is_module_installed(qt_binding)
+        for qt_binding in ('PyQt5', 'PySide2', 'PyQt6', 'PySide6')
+    ):
```
ccordoba12:
```suggestion
    for qt_binding in ('PyQt6', 'PySide6', 'PyQt5', 'PySide2'):
        if is_module_installed(qt_binding):
            auto_backend = 'qt'
            break
```

* Use `for` and `break` to avoid checking for all backends, which is inefficient (and these are heavy modules to import).
* Check for Qt6 bindings first because (eventually) they'll be the default ones.

ccordoba12 (commented):
Thanks for your help with this @larsoner!

ccordoba12:
After applying my suggestion, please rebase/merge with master to get the latest fixes to our tests.

larsoner:
Okay done @ccordoba12 ! Workflows will need manual approval

ccordoba12 (approved):
Looks good to me now, thanks @larsoner!

PR #604, opened 2026-08-24
PR: Prevent error when running `test_dask_multiprocessing`

ccordoba12:
- That's done by pinning Bokeh to a specific version, which is needed to work around an incompatibility between Bokeh and Dask.
- We don't depend on Bokeh directly but otherwise our test suite will fail.


ccordoba12:

